PR #5, opened 2020-07-07
[Config] Clean up preproduction testing code

vguerci:
Remove that code from production logic to tests only
JIRA: EE-1001

### DIFF
--- a/CriteoPublisherSdk/Sources/Configuration/CR_Config.m
+++ b/CriteoPublisherSdk/Sources/Configuration/CR_Config.m
@@ -27,19 +27,6 @@ NSString *const CR_ConfigCdbUrl = @"https://bidder.criteo.com";
 NSString *const CR_ConfigAppEventsUrl = @"https://gum.criteo.com/appevent/v1";
 NSString *const CR_ConfigConfigurationUrl = @"https://pub-sdk-cfg.criteo.com/v2.0/api/config";
 
-// FIXME EE-1001
-// Test (using Pre-Production or Local)
-// #define HIT_LOCAL_CDB // Uncomment to use local CDB instead of preprod
-#ifdef HIT_LOCAL_CDB
-// This is the default configuration if you follow the CDB readme.
-NSString *const CR_ConfigTestCdbUrl = @"http://127.0.0.1:9991";
-#else
-NSString *const CR_ConfigTestCdbUrl = @"https://directbidder-test-app.par.preprod.crto.in";
-#endif
-#undef HIT_LOCAL_CDB
-NSString *const CR_ConfigTestAppEventsUrl = @"https://pub-sdk-cfg.par.preprod.crto.in";
-NSString *const CR_ConfigTestConfigurationUrl = @"https://gum.par.preprod.crto.in";
-
 @interface CR_Config ()
 
 @property(nonatomic, strong, readonly) NSUserDefaults *userDefaults;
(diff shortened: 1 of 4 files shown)


CriteoPublisherSdk/Tests/IntegrationTests/CR_WebApiIntegrationTests.m
```diff
@@ -52,17 +52,13 @@ @implementation CR_WebApiIntegrationTests
 - (void)setUp {
   NSString *userDefaultsId = NSStringFromClass(self.class);
   [[NSUserDefaults new] removePersistentDomainForName:userDefaultsId];
```
TurpIF, line 54: This line seems to clean the state of the user default created with the userDefaultsId.
After your changes, the user defaults are created by the dependency provider, and is not using this userDefaultsId anymore.
I think that you should either remove this cleaning if it is not needed anymore.
Or, this cleaning is still required, then a solution should be found to inject the proper userDefaults.

Amokrane:
LGTM as a short term fix. However, I see some potential for further improvements (FAR).

Some endpoints are scattered around the codebase: https://github.com/criteo/ios-publisher-sdk/blob/main/CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m#L141,  https://github.com/criteo/ios-publisher-sdk/blob/main/CriteoPublisherSdk/Tests/UnitTests/Network/CR_BidRequestSerializerSwiftTests.swift#L43, 

Also, there is an inconsistency between how endpoints are fetched depending on the environment, which makes it hard to follow. In particular, product environment is obtained by calling `CR_Config` methods directly, while test environment is obtained through a level of indirection (`DependencyProvider`). Why not having the `DependencyProvider` serve a `CR_Config` based on the environment, and other parameters (such as the HIT_LOCAL_CDB which you removed).  In this case, `CR_Config` could be a simple value object.

Also (not related to this PR at all or EE-1001), it seems that `CR_Config` was originally designed to hold the config state from the remote config TLA. It's a bit strange to have URLs mixed into that, there might be an opportunity to separate environment configuration (endpoints to hit depending on the environment) from remote configuration values (kill switch,  etc).

CriteoPublisherSdk/Sources/Configuration/CR_Config.m
```diff
@@ -27,19 +27,6 @@
 NSString *const CR_ConfigAppEventsUrl = @"https://gum.criteo.com/appevent/v1";
 NSString *const CR_ConfigConfigurationUrl = @"https://pub-sdk-cfg.criteo.com/v2.0/api/config";
 
-// FIXME EE-1001
-// Test (using Pre-Production or Local)
-// #define HIT_LOCAL_CDB // Uncomment to use local CDB instead of preprod
-#ifdef HIT_LOCAL_CDB
```
Amokrane, line 33: Not needed anymore? You prefer manually editing the URL, or you just don't see the need for using a local CDB?

CriteoPublisherSdk/Tests/IntegrationTests/CR_WebApiIntegrationTests.m
Amokrane: Same concern.

vguerci: Removed as it is useless now

CriteoPublisherSdk/Sources/Configuration/CR_Config.m
vguerci: I never had to use this, which is anyway on the production side code to be removed. Yes this is as simple as editing an URL on test side, and if needed it should deserve some dependencyProvider builder method, or, as you suggested a rework on the configuration code which could be improved. In any cases, FAR imo

PR #8, opened 2020-07-08
[Tests] Fix preprod dependent tests with WireMock

vguerci:
* Setup CI to start a WireMock server before testing
* Re-enable Functional Tests that were relying on preprod and fix them pointing to local WireMock Server

PR #9, opened 2020-07-09
[Tests] Fix NetworkManagerTests with WireMock

vguerci:


vguerci:
Note that this is based on #8 and only last commit is to be reviewed > 46d9cf1

CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m
```diff
@@ -138,7 +135,7 @@ - (void)skipped_testNetworkManagerPostCall {
                                nil],
           @"slots", nil];
 
-  NSURL *url = [NSURL URLWithString:@"http://directbidder-test-app.par.preprod.crto.in/"
+  NSURL *url = [NSURL URLWithString:@"https://localhost:9099/directbidder-test-app/"
```
TurpIF, line 138: This was the URL of CDB preprod. What do you think about putting this in a constant ? So if we need to change it, we can do it easily.

wiremock/mappings/directbidder-test-app.json
```diff
@@ -0,0 +1,20 @@
...
+  "metadata": {
+    "origin": "http://directbidder-test-app.par.preprod.crto.in/inapp/v1?profileId=235"
+  },
+  "request": {
+    "url": "/directbidder-test-app/inapp/v1?profileId=235",
+    "method": "POST"
+  },
+  "response": {
+    "status" : 200,
+    "body" : "{\r\n      \"slots\": [{\r\n        \"impId\": \"5f05ec4e58f360e11c914a9de6650fc4\",\r\n        \"placementId\": \"div-Test-DirectBidder\",\r\n        \"arbitrageId\": \"arbitrage_id\",\r\n        \"cpm\": \"1.00\",\r\n        \"currency\": \"EUR\",\r\n        \"width\": 300,\r\n        \"height\": 250,\r\n        \"ttl\": 0,\r\n        \"displayUrl\": \"https:\/\/directbidder-stubs.par.preprod.crto.in\/delivery\/ajs.php?width=300&height=250\"\r\n      }],\r\n      \"requestId\": \"bffa74c5-c75e-47ff-8ddf-ead9a88f74ae\"\r\n    }",
```
TurpIF, line 12: Is it possible to really have multiline strings in wiremock mapping files ? Then it could make the response more readable. Else, do you think the the \r\n coudl be skipped ?

CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m
```diff
@@ -138,7 +135,7 @@ - (void)skipped_testNetworkManagerPostCall {
...
                                     @"inapp/v1?profileId=235"];
```
TurpIF, line 139: /inapp/v1 ?

I'm afraid that this test is completely wrong :D
AFAIK the /inapp/v1 endpoint of CDB is for replying to the Publisher SDK v1 that was made out of Criteo 2~3 years ago.  This publisher SDK should use the /inapp/v2 endpoint.

What do you think about trying to migrate this test to use the v2 endpoint ? It may look FAR, but I don't think it is a good idea neither to have such kind of test that brings nothing to this SDK.
The payload might have change a little bit.

wiremock/mappings/directbidder-test-app.json
Amokrane: When I used WireMock in the past, I remember I had to do the same escaping. FWIW, I was also compressing the output (I was using: https://www.onlinejsonconvert.com/json-escape/).

Amokrane:
Minor comment about the commit message: could you please add a description that states the problem and how WireMock is fixing the problem?

Something like:
```
[Tests] Fix NetworkManagerTests with WireMock

Some tests rely on the PreProd endpoint which often fail to respond.
WireMock is used to mock the PreProd response.
```

wiremock/mappings/directbidder-test-app.json
vguerci: JSON format does not support multiline, nor comments ¯\\\_(ツ)_/¯ it could be inlined though, without \\r\\n but as @Amokrane said, it is easy to use an escape tool.

CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m
vguerci: It does not makes much sense as it is a one shot use in this test

CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m
vguerci: Done. But as commented in commit, kind of useless as this test only covers the NetworkManager post code w/o the business logic, which is tested elsewhere. This is an integration test w/o mocking the networking call.

PR #15, opened 2020-07-20
Add Release workflow

vguerci:


.github/bin/publish-binaries
```diff
@@ -0,0 +1,176 @@
...
+# and optionally the release-ID to associate them with.
+#
+# These will be stored in the environment as:
+#
+#   INPUT_ARGS
+#
+#   INPUT_RELEASEID
+#
+# Taken from https://github.com/skx/github-action-publish-binaries
+# Action that depends on Docker which works only on Linux
```
TurpIF, line 15: I guess this is why you inlined the script ? Maybe could you says this also in the commit message to explain the why of this commit.

Also, did you try to find alternative GA to publish release artifacts ? For instance, there is https://github.com/svenstaro/upload-release-action (i didn't check if it is a good one).

.github/workflows/release.yml
```diff
@@ -0,0 +1,104 @@
...
+
+    - name: Publish binaries
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        INPUT_ARGS: build/output/*.zip
+      run: |
+        .github/bin/publish-binaries
+
+  # Clone of main workflow
+  test:
```
TurpIF, line 56: Do you need to run tests when releasing ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Ruby setup
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.6
+
+    - name: Bundle cache
```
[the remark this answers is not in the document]
vguerci: no longer relevant as Release is now just an extension of test job, in the future we should be capable of reusing jobs

.github/bin/publish-binaries
vguerci: Switched to [another one](https://github.com/softprops/action-gh-release) that is slightly more followed / better as it creates release on tag creation

.github/workflows/release.yml
```diff
@@ -0,0 +1,75 @@
...
+      uses: actions/setup-java@v1
+      with:
+        java-version: '11'
+    - name: WireMock setup
+      run: |
+        curl https://repo1.maven.org/maven2/com/github/tomakehurst/wiremock-jre8-standalone/2.27.0/wiremock-jre8-standalone-2.27.0.jar -o wiremock.jar -s
+        bundle exec iostrust add ./wiremock/cert/wiremock.crt
+        java -jar wiremock.jar --https-port 9099 --root-dir wiremock --https-keystore wiremock/cert/wiremock.jks --keystore-password password --verbose &
+
+    - name: Test
```
TurpIF, line 55: Do you think it is necessary to test just before releasing ?
Normally, everything on the main branch should be releasable. And if there is a flaky tests that turn red, then the release will be abort and we would need to delete the remote tag and re-push it.

Please confirm if you're confident on tests, I will +2 then.

vguerci: Not all commits are merged / pushed on main and therefore you cannot guarantee all will succeed at tests. For instance when you rebase / merge, only the last commit is tested. In any cases, It can't hurt to test a last time before releasing.
I'm confident on tests, and anyway I think you can retriever workflow manually if needed.

PR #20, opened 2020-07-21
Mopub adapter

vguerci:


TurpIF:
This is a squash version of the mopub adapter. Do you think that it may be worth to merge the two git histories (see --allow-unrelated-histories) to keep the trace and the recognition of authors ?

vguerci:
This is a draft, don't take much time reviewing this ;)
About history, we are talking about dozen commits with authors already in the main repo, not convinced that is worth to keep, but will look into.

CriteoMoPubAdapter/Tests/CriteoMoPubAdapterTests/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>BNDL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0</string>
+	<string>4.0.0-alpha1</string>
```
TurpIF, line 18: Shouldn't this be 4.0.0.0-alpha1 ? Or do you want to to have a unique version for all SDK+adapters ?

vguerci: With that merge there will be only one package with one version

vguerci: Also means one changelog, if you want to amend an adapter you have to push the Sdk as well

vguerci: TODO: Changelog 😅

PR #22, opened 2020-07-22
Merge mopub subtree

vguerci:


vguerci:
Not sure how Github will behave when merging this PR with a subtree, I might have to force push this if that does not work ¯\_(ツ)_/¯

TurpIF:
I do not understand your last commit (the git subtree).
The history seems to be listed, but the when you blame a file in the last commit, all changes are attached to this commit: https://github.com/criteo/ios-publisher-sdk/blame/c8f85f4b2c5b167d6285f6fb752895e382b9e4bc/CriteoMoPubAdapter/.gitignore

I don't know about the git subtree tool, do you mind checking if a simple git merge work ? Note that you may need to add a commit to prepare the merge: moving all the files of the adapter in the target directory.

TurpIF:
Nevermind, I checkout this PR to check it more deeply, and in fact .gitignore is really a 100% added file.
Git blame on old files works: https://github.com/criteo/ios-publisher-sdk/blame/c8f85f4b2c5b167d6285f6fb752895e382b9e4bc/CriteoMoPubAdapter/CriteoMoPubAdapter/CRBannerCustomEvent.m

Although I don't understand why some files are new. Did you do any preparation in the same commit ? If ou don't mind could you split the preparation commit and the merge commit ?

vguerci:
`git-subtree` is a git builtin helpful to merge two repos as a subtree, I thought it would keep the history `--follow`able. Which [is not the case](https://github.com/criteo/ios-publisher-sdk/commits/c8f85f4b2c5b167d6285f6fb752895e382b9e4bc/CriteoMoPubAdapter/CriteoMoPubAdapter/CRBannerCustomEvent.m).
I will try to do another PR with the approach you suggested. Also did a diff, I have found no new file, this is a pristine copy of the current repository, can you elaborate?

vguerci:
Abandoned to the favor of #25

TurpIF:
Don't forget to kill the branch ;)

PR #30, opened 2020-07-23
Google adapter integration

vguerci:


vguerci:
Added the Adapters READMEs to #32

PR #33, opened 2020-07-24
Profile registry

vguerci:


vguerci:
Because I need your changes to continue, I always hope to avoid some rebase / ci cycle by doing so, but maybe that's not necessary...

TurpIF:
This seems pretty orthogonal though

PR #36, opened 2020-07-24
Implement Integration Registry

TurpIF:
- [x] Declared integration is persisted across sessions
- [x] Provide a method to retrieve profile ID
- [x] Mediation adapters are detected
- [x] If mediation are detected, mediation is kept even if standalone declare itself
- [x] If no info is available (for prefetch, remote config) use fallback profile ID = 235

JIRA: EE-1196

### DIFF
--- a/CriteoPublisherSdk/Sources/Configuration/CR_IntegrationRegistry.m
+++ b/CriteoPublisherSdk/Sources/Configuration/CR_IntegrationRegistry.m
@@ -19,6 +19,8 @@
 
 #import "CR_IntegrationRegistry.h"
 
+NSString *const NSUserDefaultsIntegrationKey = @"CRITEO_ProfileId";
+
 @interface CR_IntegrationRegistry ()
 
 @property(nonatomic, strong, readonly) NSUserDefaults *userDefaults;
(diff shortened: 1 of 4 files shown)


CriteoPublisherSdk/Sources/Configuration/CR_IntegrationRegistry.m
```diff
@@ -45,6 +47,12 @@ - (void)declare:(CR_IntegrationType)integrationType {
 }
 
 - (NSNumber *)profileId {
+  if (self.isMoPubMediationPresent) {
+    return @(CR_IntegrationMopubMediation);
+  } else if (self.isAdMobMediationPresent) {
```
vguerci, line 52: Does this have a strong requirement of mutually exclusive?
I'm thinking of the test app, that could help not to be.

TurpIF: What do you propose when both are detected ?

vguerci: Well, I doubt we want to introduce bit mask logic there to declare both at the same time.
What bothers me is that in such situation Google "wins".
On test app we will override the value as you did on Android.
Maybe if both are detected we should report `Fallback` instead?

TurpIF: Ok to use fallback then, it makes more sense

PR #38, opened 2020-07-24
[github] Tests reports

vguerci:


.github/workflows/main.yml
```diff
@@ -60,3 +60,16 @@ jobs:
     - name: Test
       run: |
         bundle exec fastlane test
+
+    - name: Test reports upload
+      uses: actions/upload-artifact@master
```
TurpIF, line 65: upload-artifact is now using main ad default branch

.github/workflows/main.yml
```diff
@@ -60,3 +60,16 @@ jobs:
...
+      with:
+        name: test-reports
+        path: "fastlane/test_output"
+
+    - name: Tests reports annotations
+      uses: ashley-taylor/junit-report-annotations-action@1.3
```
TurpIF, line 71: FYI this action is currently broken. This [PR](https://github.com/ashley-taylor/junit-report-annotations-action/pull/16) should fix it.

vguerci:
Removed annotations until It works, fixed the upload version.
Side note, you should avoid targeting @main / @master as it is fragile and could even be, even if unlikely, an attack vector.

vguerci:
Switched to main as it is not v4 specific _(and also because tests are failing on v4 🤔)_

PR #39, opened 2020-07-27
Add test validating that adapters are detected

TurpIF:
Adapters are detected via introspection in #36. Those tests guarantee that this weak link works.

JIRA: EE-1196

### DIFF
--- a/CriteoGoogleAdapter/CriteoGoogleAdapter.xcodeproj/project.pbxproj
+++ b/CriteoGoogleAdapter/CriteoGoogleAdapter.xcodeproj/project.pbxproj
@@ -12,6 +12,7 @@
 		54FC7DBC22F4147100202164 /* CRInterstitialCustomEvent.h in Headers */ = {isa = PBXBuildFile; fileRef = 54FC7DBA22F4147100202164 /* CRInterstitialCustomEvent.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		54FC7DBD22F4147100202164 /* CRInterstitialCustomEvent.m in Sources */ = {isa = PBXBuildFile; fileRef = 54FC7DBB22F4147100202164 /* CRInterstitialCustomEvent.m */; };
 		A36837911ADA01CEDFA729A9 /* libPods-CriteoGoogleAdapterTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = BF43AF45DE0BEECB9A9247A6 /* libPods-CriteoGoogleAdapterTests.a */; };
+		C07ABB57E4BA18C3A8A35B47 /* CR_AdMobIntegrationRegistryTests.m in Sources */ = {isa = PBXBuildFile; fileRef = C07AB1B8BF317626BD709C00 /* CR_AdMobIntegrationRegistryTests.m */; };
 		CF6AE25E24CEB22F00E00CDA /* CriteoPublisherSdk.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = CF6AE25D24CEB22F00E00CDA /* CriteoPublisherSdk.framework */; };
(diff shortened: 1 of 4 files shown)


CriteoGoogleAdapter/CriteoGoogleAdapter.xcodeproj/project.pbxproj
```diff
@@ -80,6 +83,7 @@
 		3CF2926D760C675680C3BA20 /* Frameworks */ = {
 			isa = PBXGroup;
 			children = (
+				CF6AE25D24CEB22F00E00CDA /* CriteoPublisherSdk.framework */,
```
vguerci, line 88: I doubt you have to add that framework, if you use the workspace that should be automatic.

CriteoGoogleAdapter/CriteoGoogleAdapter.xcodeproj/project.pbxproj
```diff
@@ -80,6 +83,7 @@
...
 				DFC3F4A323035EF500318644 /* CriteoPublisherSdk.framework */,
```
vguerci, line 89: FAR: This is probably a leftover of a previous version that should be removed

CriteoGoogleAdapter/CriteoGoogleAdapter.xcodeproj/project.pbxproj
TurpIF: The tests didn't pass for the AdMob adapter. I diff the AdMob/MoPub config to see were it comes from, and apparently this is a fix.

CriteoGoogleAdapter/CriteoGoogleAdapter.xcodeproj/project.pbxproj
TurpIF: Good catch ! I rebased to switch the two commits. And merging pbxproj is completely dumb :(

TurpIF:
> This is weak and fragile, hopefully not a big deal now that tests are all run at once.
> FAR: What about adding end to end tests with wiremock to check the profile ids on different scenarios?

The detection is also weak :D When using reflection, I prefer to protect the code via unit tests. Here the tests also use reflection... But at least if those breaks, the tests are quite small to be easily debugged and fixed.

Big +1 for E2E tests. Maybe not down to wiremock, maybe we could just check profile ID given to NetworkManager. However, this is FAR as it would need the whole feature to be merged. Let me add a ticket for this.

CriteoGoogleAdapter/CriteoGoogleAdapter.xcodeproj/project.pbxproj
TurpIF: Done

CriteoGoogleAdapter/CriteoGoogleAdapter.xcodeproj/project.pbxproj
vguerci: Interesting, that is only because of the use of `NSProtocolFromString`, that class is discarded not being explicitly referenced otherwise.

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Criteo Publisher SDK Changelog
+
+### Features
+- Consider safe area when displaying interstitial Ads for Standalone, In-House, MoPub AppBidding and
```
vguerci, line 4: Sounds more like a bug fix than a feature from a publisher's perspective

TurpIF: Ok, but this is part of the merge

vguerci:
Now that v4 has been merged into main, you should target main as it is not v4 specific

TurpIF:
Cool, thank you, I will do this

PR #42, opened 2020-07-27
[profile-id] Integrate Registry

vguerci:
JIRA: EE-1197

### DIFF
--- a/CriteoPublisherSdk/Sources/Feedback/CR_FeedbackController.m
+++ b/CriteoPublisherSdk/Sources/Feedback/CR_FeedbackController.m
@@ -25,6 +25,7 @@
 #import "CR_ApiHandler.h"
 #import "Logging.h"
 #import "CR_FeedbackFeatureGuard.h"
+#import "CR_IntegrationRegistry.h"
 
 @interface CR_FeedbackController ()
 
@@ -69,8 +70,9 @@
 - (void)onCdbCallStarted:(CR_CdbRequest *)request {
   NSString *requestGroupId = [CR_UniqueIdGenerator generateId];
   for (NSString *impressionId in request.impressionIds) {
-    [self.feedbackStorage setCdbStartAndImpressionIdForImpressionId:impressionId
-                                                     requestGroupId:requestGroupId];
+    [self.feedbackStorage setCdbStartForImpressionId:impressionId
+                                           profileId:request.profileId
+                                      requestGroupId:requestGroupId];
   }
 }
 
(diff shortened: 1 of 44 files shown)


CriteoPublisherSdk/Sources/Network/Serializers/CR_BidRequestSerializer.m
```diff
@@ -48,8 +45,8 @@ - (instancetype)initWithGdprSerializer:(CR_GdprSerializer *)gdprSerializer {
 
 #pragma mark - Public
 
-- (NSURL *)urlWithConfig:(CR_Config *)config {
-  NSString *query = [NSString stringWithFormat:@"profileId=%@", config.profileId];
+- (NSURL *)urlWithProfileId:(NSNumber *)profileId config:(CR_Config *)config {
```
TurpIF, line 48: Profile ID is useless in the query string. What do you think about taking the opportunity to remove it ? It may seems FAR but I think that injecting the new profile id there add too many useless stuff.

See the [schema on CDB](http://review.crto.in/gitweb?p=publisher/direct-bidder.git;a=blob;f=directbidder-app/src/main/scala/com/criteo/directbidder/inapp/InApp.scala;h=4e38f9e8da2e0d5ce8c78d8430313466acb0710c;hb=HEAD#l37)

CriteoPublisherSdk/Sources/Feedback/CR_FeedbackController.m
```diff
@@ -111,18 +113,36 @@ - (void)onBidConsumed:(CR_CdbBid *)consumedBid {
...
     return;
   }
 
+  NSDictionary<NSNumber *, NSArray<CR_FeedbackMessage *> *> *messagesByProfileId =
+      [messages cr_groupByKey:^NSNumber *(CR_FeedbackMessage *message) {
+        return message.profileId ?: @(CR_IntegrationFallback);
+      }];
+  [messagesByProfileId
+      enumerateKeysAndObjectsUsingBlock:^(NSNumber *profileId,
+                                          NSArray<CR_FeedbackMessage *> *messages, BOOL *stop) {
```
TurpIF, line 127: this messages variable is shadowing the one above. What do you think about using a more meaningful name for this one ?

CriteoPublisherSdk/Sources/Feedback/CR_FeedbackController.m
```diff
@@ -111,18 +113,36 @@ - (void)onBidConsumed:(CR_CdbBid *)consumedBid {
...
+        [self sendFeedbacks:messages profileId:profileId];
+      }];
+}
+
+- (void)sendFeedbacks:(NSArray<CR_FeedbackMessage *> *)messages profileId:(NSNumber *)profileId {
+  if (messages.count == 0) {
```
TurpIF, line 133: Is this condition possible ?
Maybe even without both conditions (this one and the one at line 117), this is working fine as you will loop on an empty dictionnary

CriteoPublisherSdk/Sources/Network/Serializers/CR_BidRequestSerializer.m
vguerci: Done.

CriteoPublisherSdk/Sources/Feedback/CR_FeedbackController.m
vguerci: Moving that precondition to the underlying ApiHandler method, where it makes sense.

PR #46, opened 2020-07-27
Integrations declare themselves to the registry

TurpIF:
- [x] Standalone:
  - `CRBannerView#loadAd`
  - `CRInterstitial#loadAdCRNativeLoader#loadAd`
  - `CRNativeLoader#loadAd`
- [x] InHouse: `Criteo#getBidResponseForAdUnit:CRAdUnit`
- [x] MoPub App Bidding: `Criteo#setBidsForRequest:id withAdUnit:CRAdUnit` with and id of type MPAdView or MPInterstitialAdController
- [x] DFP App Bidding: `Criteo#setBidsForRequest:id withAdUnit:CRAdUnit` with and id of type DFPRequest (or siblings)
- [x] Custom App Bidding: `Criteo#setBidsForRequest:id withAdUnit:CRAdUnit` with and id of type NSMutableDictionary

JIRA: EE-1196

### DIFF
--- a/CriteoPublisherSdk/Tests/IntegrationTests/CRInterstitialTests.m
+++ b/CriteoPublisherSdk/Tests/IntegrationTests/CRInterstitialTests.m
@@ -39,84 +39,46 @@
 #import "CR_DependencyProvider+Testing.h"
 #import "CR_DisplaySizeInjector.h"
 
-@interface CRInterstitialTests : XCTestCase {
-  CR_CdbBid *_bid;
-  CR_CacheAdUnit *_cacheAdUnit;
-  CRInterstitialAdUnit *_adUnit;
-}
+@interface CRInterstitialTests : XCTestCase
+
+@property(nonatomic, strong) CR_CacheAdUnit *expectedCacheAdUnit;
+@property(nonatomic, strong) CRInterstitialAdUnit *adUnit;
+@property(nonatomic, strong) Criteo *criteo;
+@property(nonatomic, strong) CR_URLOpenerMock *urlOpener;
+@property(nonatomic, strong) CR_DisplaySizeInjector *displaySizeInjector;
+
 @end
 
 @implementation CRInterstitialTests
 
 - (void)setUp {
-  _bid = nil;
-  _cacheAdUnit = nil;
-  _adUnit = nil;
-}
+  self.expectedCacheAdUnit = [[CR_CacheAdUnit alloc] initWithAdUnitId:@"123"
+                                                                 size:CGSizeMake(320.0, 480.0)
+                                                           adUnitType:CRAdUnitTypeInterstitial];
+  self.adUnit = [[CRInterstitialAdUnit alloc] initWithAdUnitId:@"123"];
 
-- (CR_CdbBid *)bidWithDisplayURL:(NSString *)displayURL {
-  if (!_bid) {
-    _bid = [[CR_CdbBid alloc] initWithZoneId:@123
-                                 placementId:@"placementId"
-                                         cpm:@"4.2"
-                                    currency:@"₹😀"
-                                       width:@47.0f
-                                      height:[NSNumber numberWithFloat:57.0f]
-                                         ttl:26
-                                    creative:@"THIS IS USELESS LEGACY"
-                                  displayUrl:displayURL
-                                  insertTime:[NSDate date]
-                                nativeAssets:nil
-                                impressionId:nil];
-  }
-  return _bid;
-}
+  self.urlOpener = CR_URLOpenerMock.new;
 
-- (CR_CacheAdUnit *)expectedCacheAdUnit {
-  if (!_cacheAdUnit) {
-    _cacheAdUnit = [[CR_CacheAdUnit alloc] initWithAdUnitId:@"123"
-                                                       size:CGSizeMake(320.0, 480.0)
-                                                 adUnitType:CRAdUnitTypeInterstitial];
-  }
-  return _cacheAdUnit;
-}
+  CR_DependencyProvider *dependencyProvider = CR_DependencyProvider.testing_dependencyProvider;
 
-- (CRInterstitialAdUnit *)adUnit {
-  if (!_adUnit) {
-    _adUnit = [[CRInterstitialAdUnit alloc] initWithAdUnitId:@"123"];
-  }
-  return _adUnit;
+  self.displaySizeInjector = OCMClassMock([CR_DisplaySizeInjector class]);
+  dependencyProvider.displaySizeInjector = self.displaySizeInjector;
+
+  self.criteo = OCMPartialMock([Criteo.alloc initWithDependencyProvider:dependencyProvider]);
 }
 
 - (void)testInterstitialSuccess {
-  CR_DependencyProvider *dependencyProvider = CR_DependencyProvider.testing_dependencyProvider;
-
-  Criteo *mockCriteo = OCMStrictClassMock([Criteo class]);
-  OCMStub(mockCriteo.config).andReturn([[CR_Config alloc] initWithCriteoPublisherId:@"123"]);
-  OCMStub(mockCriteo.dependencyProvider).andReturn(dependencyProvider);
-
   MockWKWebView *mockWebView = [MockWKWebView new];
-  CR_InterstitialViewController *interstitialVC =
-      [[CR_InterstitialViewController alloc] initWithWebView:mockWebView view:nil interstitial:nil];
-  CRInterstitial *interstitial =
-      [[CRInterstitial alloc] initWithCriteo:mockCriteo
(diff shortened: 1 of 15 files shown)


CriteoPublisherSdk/Tests/UnitTests/CriteoTests.m
```diff
@@ -0,0 +1,55 @@
...
+#import "CR_DependencyProvider+Testing.h"
+
+@interface CriteoTests : XCTestCase
+
+@property(strong, nonatomic) Criteo *criteo;
+@property(strong, nonatomic) CR_IntegrationRegistry *integrationRegistry;
+
+@end
+
+@implementation CriteoTests
```
vguerci, line 36: That is interesting that this test class is just introduced now. That reveals that:
- this `Criteo` class is under tested
- but also that its scope is unclear, maybe it should not be responsible of such in house code

FAR: Some refactoring & tests

CriteoPublisherSdk/Tests/UnitTests/Standalone/CRBannerViewTests.m
```diff
@@ -32,123 +32,75 @@
 #import "NSURL+Criteo.h"
 #import "CR_TokenValue+Testing.h"
 #import "XCTestCase+Criteo.h"
+#import "CR_DependencyProvider.h"
```
vguerci, line 35: This import is not necessary as it comes from the one after, as it is generally the case with extensions:
https://github.com/criteo/ios-publisher-sdk/blob/60713fe8d1784550afb2333b1841a0854ddbb0f7/CriteoPublisherSdk/Tests/Utility/Categories/CR_DependencyProvider%2BTesting.h#L20

CriteoPublisherSdk/Tests/UnitTests/Standalone/CRBannerViewTests.m
```diff
@@ -32,123 +32,75 @@
...
+  self.criteo = OCMPartialMock([Criteo.alloc initWithDependencyProvider:dependencyProvider]);
 }
 
-- (CRBannerAdUnit *)adUnit {
-  if (!_adUnit) {
-    _adUnit = [[CRBannerAdUnit alloc] initWithAdUnitId:@"123" size:CGSizeMake(47.0, 57.0)];
-  }
-  return _adUnit;
+- (void)tearDown {
+  self.criteo = nil;
```
vguerci, line 70: Is this really required?

CriteoPublisherSdk/Tests/UnitTests/Standalone/CRBannerViewTests.m
```diff
@@ -32,123 +32,75 @@
...
-  OCMStub(mockCriteo.config).andReturn([[CR_Config alloc] initWithCriteoPublisherId:@"123"]);
-  MockWKWebView *mockWebView = [MockWKWebView new];
-
-  CRBannerView *bannerView =
-      [[CRBannerView alloc] initWithFrame:CGRectMake(13.0f, 17.0f, 47.0f, 57.0f)
-                                   criteo:mockCriteo
-                                  webView:mockWebView
-                                   adUnit:self.adUnit
-                                urlOpener:self.urlOpener];
+  MockWKWebView *webView = [MockWKWebView new];
```
vguerci, line 74: NIT `mockWebView`

CriteoPublisherSdk/Tests/IntegrationTests/CRInterstitialTests.m
```diff
@@ -59,10 +61,12 @@ - (void)setUp {
 
   self.urlOpener = CR_URLOpenerMock.new;
 
-  CR_DependencyProvider *dependencyProvider = CR_DependencyProvider.testing_dependencyProvider;
-
   self.displaySizeInjector = OCMClassMock([CR_DisplaySizeInjector class]);
+  self.integrationRegistry = OCMClassMock([CR_IntegrationRegistry class]);
```
vguerci, line 65: FAR: Would make sense to do this mocking by default in DependencyProvider+Testing

CriteoPublisherSdk/Tests/UnitTests/CriteoTests.m
TurpIF: Agree it is definitely under tested AND responsibility is not well defined here.
On Android, we made appear an InHouse class. This make the Criteo only a facade having the API responsibility only.

CriteoPublisherSdk/Tests/UnitTests/Standalone/CRBannerViewTests.m
TurpIF: Long story short: another painful testRendering (remember the ultra flaky one ??)

I do not know why but there is a hidden side effect somewhere that break this test even if the body of this test remain unchanged. Setting criteo to nil (although it is setup to new instance every time) seems to fix the test.

I tried to do the same on the testRendering from interstitial with no chance :/

CriteoPublisherSdk/Tests/IntegrationTests/CRInterstitialTests.m
TurpIF: The issue is that some tests would want to have mocks while other would want real instances or even spies.
On Android we make this easier with annotations and injection via a junit rule (https://github.com/criteo/android-publisher-sdk/blob/main/publisher-sdk-tests/src/androidTest/java/com/criteo/publisher/integration/IntegrationRegistryTest.kt#L52 for instance).

Maybe we could find a way to make this nicer in obj-c too ?

TurpIF:
Done, I fixed the your nit comments

CriteoPublisherSdk/Tests/IntegrationTests/CRInterstitialTests.m
vguerci: In that specific case, the registry should be tested directly with a unit test, when it comes to integration tests it does not make sense and should always be mocked, thus the default init, mock that could be overrided in a specific test anyway. The Java annotations are syntactic sugar for initializations, but the concept is the same, isn't?
Anyway FAR

CriteoPublisherSdk/Tests/UnitTests/Standalone/CRBannerViewTests.m
vguerci: I think that would deserve some comment

CriteoPublisherSdk/Tests/UnitTests/CriteoTests.m
vguerci: Let's introduce that class then and remove that test class then?

CriteoPublisherSdk/Tests/UnitTests/Standalone/CRBannerViewTests.m
vguerci: On second thoughts, that does not makes sense as this property should be deallocated anyway, TestCases are allocated at each test run, and therefore dealloc is implicit after. Are you sure this has effects?

vguerci: Now I remember we has seen similar things in the past with Romain, will ask him, I suspect there is a bug or some kind of subtlety in [OCMock partial mocks](https://ocmock.org/reference/#partial-mocks) they state that it should just works, but it apparently does not.

CriteoPublisherSdk/Tests/IntegrationTests/CRInterstitialTests.m
TurpIF: Ok, but that may be good to have a solution that fit both integration and unit tests :D

About the Java annotations, no it's just annotation acting as a marker. Then there is an annotation processor looking for annotated elements and injecting mock/spy/instance in fields and the dependency provider: https://github.com/criteo/android-publisher-sdk/blob/main/test-utils/src/main/java/com/criteo/publisher/mock/DependenciesAnnotationInjection.java

CriteoPublisherSdk/Tests/UnitTests/Standalone/CRBannerViewTests.m
TurpIF: image

I will add a comment

TurpIF: done

CriteoPublisherSdk/Tests/UnitTests/CriteoTests.m
TurpIF: Isn't this FAR ?

vguerci: FAR is fine as well ¯\\\_(ツ)_/¯

PR #50, opened 2020-07-29
[profile-id] Tests fixes

vguerci:
The Feedback functional tests were broken because of profile id introduced changes.
Let's isolate most tests from integration registry.


TurpIF:
image

vguerci:
Waiting for #42 to pass tests before merging, might be some additional work to do there

vguerci:
Side note, I'm 90% confident we have flaky tests (Interstitial test again) in the opposite of the usual way, i.e. bringing false positive results ¯\\\_(ツ)_/¯

PR #51, opened 2020-07-29
[tools][clang-format] Add git hook

vguerci:


scripts/dev-setup.sh
```diff
@@ -11,11 +11,15 @@ xcrun simctl spawn booted log config --mode "level:off"  --subsystem com.apple.C
 
 # Install Azure CLI with Homebrew for releasing
 # https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-macos?view=azure-cli-latest
-brew update && brew install azure-cli
+brew install azure-cli
```
TurpIF, line 14: Azure is still used ?

vguerci: Nope, but until releasing 3.9 there is no harm keeping this. I have a more wide cleanup task in my TODO list that includes this ;)

PR #57, opened 2020-07-30
Release fix

vguerci:


.github/workflows/release-cocoapods.yml
```diff
@@ -0,0 +1,38 @@
+name: Release on CocoaPods
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
```
TurpIF, line 6: Do you really want a workflow_dispatch for releasing ?

.github/workflows/release-cocoapods.yml
```diff
@@ -0,0 +1,38 @@
...
+          ${{ runner.os }}-gems-
+    - name: Bundle install
+      run: |
+        gem install bundler
+        bundle config path vendor/bundle
+        bundle install --jobs 4 --retry 3
+
+    - name: CocoaPods Release
+      run: |
+        bundle exec fastlane
```
TurpIF, line 36: bundle exec fastlane release_cocoapods ?

fastlane/Fastfile
```diff
@@ -102,6 +102,9 @@ platform :ios do
 
   desc %(Archive then release version to GitHub and CocoaPods)
   lane :release do |options|
+    if ref = options[:ref]
+      options[:version] = ref.split("/").last
```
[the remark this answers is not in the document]
vguerci: Done, also added filters on tag so release isn't triggered on non semver branches

fastlane/Fastfile
```diff
@@ -105,7 +105,6 @@ platform :ios do
     if ref = options[:ref]
       options[:version] = ref.split("/").last
     end
-    archive
```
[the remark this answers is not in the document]
vguerci: This works locally but for some reason not on GitHub, I'd like to fix this someday to be GitHub agnostic as possible

.github/workflows/release-cocoapods.yml
vguerci: This is handy to test workflow, as publishing a release is not the easiest thing to do. Anyway, removed.

.github/workflows/release-cocoapods.yml
vguerci: Fixed

vguerci:
Yup but notice 40m before another one only to rebase to prevent noise in the later ;)

PR #68, opened 2020-08-26
WireMock Fix

vguerci:
By default, x509 certs have 30 days, not considering cnf default_days option. That should fix the broken tests.


TurpIF (approved):
Thank you for fixing the tests.

I wonder if those tests will start failing again in 10 years 😄

PR #70, opened 2020-08-27
[AdViewer] Integrate into project

vguerci:
- Cleanup deprecated legacy files, scripts...
- Rework file structure in a swift package manner, as other projects
- **Xcode**: Integrate into workspace (dev workflow simplified: edit/build sdk, run test app)
- **Fastlane**: test, format, versioning (means CI now checks Test app builds & starts)


Amokrane (changes requested):
Can you add the licensing headers like we did on the SDK?

vguerci:
> Can you add the licensing headers like we did on the SDK?

Done in 4f70889

Amokrane (approved):

PR #75, opened 2020-09-02
[Dependencies] Update Gems and Pods

vguerci:


Gemfile.lock
```diff
@@ -14,16 +14,16 @@ GEM
       json (>= 1.5.1)
     atomos (0.1.3)
     aws-eventstream (1.1.0)
-    aws-partitions (1.358.0)
-    aws-sdk-core (3.104.4)
+    aws-partitions (1.363.0)
+    aws-sdk-core (3.105.0)
```
Amokrane, line 18: What do we use AWS libraries for?

Amokrane (approved):

Gemfile.lock
vguerci: we don't, these are Fastlane dependencies, which has a [s3](https://docs.fastlane.tools/actions/s3/) feature

PR #76, opened 2020-09-02
[iOS14] App Install capacity check

vguerci:
JIRA: EE-1254

### DIFF
--- a/CriteoAdViewer/Sources/Info.plist
+++ b/CriteoAdViewer/Sources/Info.plist
@@ -50,5 +50,14 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>NSUserTrackingUsageDescription</key>
+	<string>Your data will be used to deliver personalized ads to you.</string>
+	<key>SKAdNetworkItems</key>
+	<array>
+		<dict>
+			<key>SKAdNetworkIdentifier</key>
+			<string>hs6bdukanm.skadnetwork</string>
+		</dict>
+	</array>
 </dict>
 </plist>
--- a/CriteoPublisherSdk/Sources/CRConstants.h
+++ b/CriteoPublisherSdk/Sources/CRConstants.h
@@ -22,4 +22,6 @@
 
 #define CRITEO_PUBLISHER_SDK_VERSION @"3.10.0"
 
+#define CRITEO_AD_NETWORK_IDENTIFIER @"hs6bdukanm.skadnetwork"
+
 #endif /* CRConstants_h */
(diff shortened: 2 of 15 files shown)


vguerci:
Not ready to review as implementation might change

CriteoPublisherSdk/Tests/Utility/Categories/NSString+APIKeys.m
```diff
@@ -75,6 +75,10 @@ + (NSString *)deviceIdTypeValue {
   return @"IDFA";
 }
 
++ (NSString *)isAppInstallCapable {
```
Amokrane, line 78: I shared some feedbacks regarding this API internally.

Amokrane (approved):

Amokrane (changes requested):
Mis-clicked the first time, sorry.

CriteoPublisherSdk/Sources/CRConstants.h
```diff
@@ -22,4 +22,6 @@
 
 #define CRITEO_PUBLISHER_SDK_VERSION @"3.10.0"
 
+#define CRITEO_AD_NETWORK_IDENTIFIER @"hs6bdukanm.skadnetwork"
```
Amokrane, line 25: Emphasizing one of the recommendations that I made internally: should we have this in a CaC or returned by CDB instead? If we revoke this ID and generate a new one (who knows for what reason), this would prevent us from having to update the SDK.

Amokrane: In fact, following up on my API feedback, I think the SDK shouldn't even know about this ID (except may be to define it in the `Info.plist`). The SDK should send the full list of IDs defined in the `Info.plist` to the CDB as the campaign should be attributed to either of the DSPs represented by that ID.

CriteoAdViewer/Sources/Info.plist
```diff
@@ -50,5 +50,14 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>NSUserTrackingUsageDescription</key>
+	<string>Your data will be used to deliver personalized ads to you.</string>
+	<key>SKAdNetworkItems</key>
+	<array>
+		<dict>
+			<key>SKAdNetworkIdentifier</key>
+			<string>hs6bdukanm.skadnetwork</string>
```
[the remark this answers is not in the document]
vguerci: No, this is just the Test App Info.plist. Also unlike on maven / Gradle, there is no such resources bundling and merging logic with frameworks. That has to be explicitly done by publisher on its app, as described in the Apple and [Criteo](https://publisherdocs.criteotilt.com/app/ios/ios14/) documentation.

vguerci:
Thanks for the extensive review but unfortunately we might have crossed each other's path as I was marking it as Draft with [a comment not to review it](https://github.com/criteo/ios-publisher-sdk/pull/76#issuecomment-686508151).

PR #78, opened 2020-09-07
IAB-TCF-V2-Objective-C lib import

vguerci:
Let's use [IAB-TCF-V2-Objective-C](https://github.com/Singlespot/IAB-TCF-V2-Objective-C) to parse TCF v2 `TCString` format.


Amokrane (changes requested):
Not needed anymore, I suppose.

vguerci:
Nope, playing with a CMP sdk confirms this, closing.

PR #80, opened 2020-09-09
Integration test: Bid Caching / "Immediate" bid (CPM > 0, TTL = 0) 

vguerci:
Also refactor the code a bit to improve code style and testability

### DIFF
--- a/CriteoPublisherSdk/Sources/CR_CdbBid.h
+++ b/CriteoPublisherSdk/Sources/CR_CdbBid.h
@@ -46,6 +46,11 @@
  */
 @property(assign, nonatomic, readonly) BOOL isRenewable;
 
+/**
+ * YES if a bid is "immediate". i.e. CPM > 0, TTL = 0
+ */
+@property(assign, nonatomic, readonly) BOOL isImmediate;
+
 + (instancetype)emptyBid;
 
 /*
@@ -73,6 +78,8 @@
 - (BOOL)isEmpty;
 - (BOOL)isExpired;
 - (BOOL)isValid;
+
+- (void)setDefaultTtl;
 
 @end
 
--- a/CriteoPublisherSdk/Sources/CR_CdbBid.m
+++ b/CriteoPublisherSdk/Sources/CR_CdbBid.m
@@ -21,6 +21,7 @@
 #import "Logging.h"
 #import "NSObject+Criteo.h"
 #import "NSString+Criteo.h"
+#import "CRConstants.h"
 
 @interface CR_CdbBid ()
 
@@ -81,7 +82,6 @@ static CR_CdbBid *emptyBid;
 }
 
 - (instancetype)initWithDict:(NSDictionary *)slot receivedAt:(NSDate *)receivedAt {
-  const double defaultValue = 900;
   self = [super init];
   if (self) {
     NSNumber *zoneId = slot[@"zoneId"];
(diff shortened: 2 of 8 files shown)


CriteoPublisherSdk/Sources/CR_CdbBid.h
```diff
@@ -74,6 +79,8 @@
 - (BOOL)isExpired;
 - (BOOL)isValid;
 
+- (void)setDefaultTtl;
```
Amokrane, line 82: I suggest adding the time unit to the method name.

CriteoPublisherSdk/Sources/CR_CdbBid.m
```diff
@@ -95,12 +94,8 @@ - (instancetype)initWithDict:(NSDictionary *)slot receivedAt:(NSDate *)receivedA
     NSNumber *height = slot[@"height"];
     NSString *creative = slot[@"creative"];
     NSString *impId = slot[@"impId"];
-    // Hard coding to 15 minutes for now
-    // TODO: move this default to the config
-    NSTimeInterval ttl = (slot && slot[@"ttl"]) ? [slot[@"ttl"] doubleValue] : defaultValue;
-    if ([cpm doubleValue] > 0 && ttl == 0) {
-      ttl = defaultValue;
-    }
+    NSTimeInterval ttl =
```
Amokrane, line 97: Same: adding the time unit to the variable name.

CriteoPublisherSdk/Sources/CR_CdbBid.m
```diff
@@ -215,13 +210,19 @@ - (BOOL)isInSilenceMode {
...
+  return self.isValidCpm && self.isValidNativeAssetsOrUrl;
 }
 
 - (BOOL)isRenewable {
-  BOOL result = !self.isInSilenceMode || self.isExpired;
-  return result;
+  return !self.isInSilenceMode || self.isExpired;
+}
+
+- (BOOL)isImmediate {
```
Amokrane, line 220: nit: could be a good occasion to document each of the `isXXX` methods to educate the reader on the meaning of these different modes.

Amokrane (changes requested):

CriteoPublisherSdk/Sources/CR_CdbBid.m
vguerci: I doubt there is any value naming the variable that way as `NSTimeInterval` type has [_by definition_](https://developer.apple.com/documentation/foundation/nstimeinterval) a second unit.

CriteoPublisherSdk/Sources/CR_CdbBid.h
vguerci: I'm not sure this is something we want to do. This method basically sets the Default TTL into in inner state, from outside caller this is regardless of the underlying unit which is an implementation detail.

CriteoPublisherSdk/Sources/CR_CdbBid.m
vguerci: Agreed but FAR, are you thinking about more documentation than the one I've put?
> YES if a bid is "immediate". i.e. CPM > 0, TTL = 0
Also, is there any reference documentation about all these modes?

CriteoPublisherSdk/Sources/CR_CdbBid.m
Amokrane: Ah ok, well it has some value for someone that doesn't know Objective-C such as myself :) But agreed that it is not necessary in this case.

CriteoPublisherSdk/Sources/CR_CdbBid.m
Amokrane: Yes I was thinking about something more explicit that explains what `immediate` means for example. But it's really nit. There is the PubSDK overview which is pinned on the Slack channel.

Amokrane (approved):

CriteoPublisherSdk/Sources/CR_CdbBid.m
vguerci: Ok will add some entry in my _personal_ todo list then :)

PR #81, opened 2020-09-09
Integration test: Bid Caching / Missing bid

vguerci:


CriteoPublisherSdk/Tests/IntegrationTests/CR_BidManagerIntegrationTests.m
```diff
@@ -121,16 +121,41 @@ - (void)test_givenPrefetchingBid_whenGetImmediateBid_ShouldPopulateCacheWithTtlS
...
+- (void)test_givenPrefetchingBid_whenNoBid_ShouldNotPopulateCache {
+  [self givenMockedCdbResponseBids:@[]];
+
+  [self.criteo testing_registerWithAdUnits:@[ self.adUnit1 ]];
+  [self.criteo testing_waitForRegisterHTTPResponses];
+
+  CR_CdbBid *cachedBid = [self.bidManager getBid:self.cacheAdUnit1];
+  XCTAssertEqualObjects(cachedBid, CR_CdbBid.emptyBid);
+
+  [self givenUnmockedCdbResponse];
```
Amokrane, line 133: Why did you include this scenario in this test? It seems like there are two scenarios here:

```
// given mocked response with nobid
// when fetching bid from cache
// assert the response is empty
```

And then:

```
// given not mocked bid responses
// when fetching bid from cache
// then assert it is not empty
```

Is this on purpose?

vguerci: Yes this is on purpose, as most scenario are described, [this particular one](https://confluence.criteois.com/pages/viewpage.action?spaceKey=PUBSDK&title=Standalone#Standalone-bannerMissingBidMissingbid) is something like:
> Then, for first click, banner is not displayed
> Then, for second click, banner is displayed

Basically most of these scenarios require that after checking one behaviour, to verify the next call one, what you perceive as a second scenario is just the continuity of the first and could be worded as:
> after getting a nobid from cache, we should prefetch another bid normally

Amokrane (approved):

Amokrane (approved):

PR #99, opened 2020-09-22
[Live Bidding] Preparation work

vguerci:


TurpIF (approved):

TurpIF:
I've noticed that you're still using the v4. As (normally) the next version is the v4, I guess you can merge it on main.

vguerci:
Ok I'll PR a v4 merge then, thank you!

PR #104, opened 2020-09-24
[Criteo] Document class

vguerci:
It would be nice at some point for public APIs to be fully documented for IDE hints or html generation.
`Criteo` class is a good start, also this PR is a way to discuss the v4 API future.

### DIFF
--- a/CriteoPublisherSdk/Sources/Criteo.h
+++ b/CriteoPublisherSdk/Sources/Criteo.h
@@ -25,22 +25,46 @@
 #import <CriteoPublisherSdk/CRBidResponse.h>
 
 NS_ASSUME_NONNULL_BEGIN
+
 @interface Criteo : NSObject
 
-/* @abstract Use sharedInstance */
+/**
+ * Use sharedCriteo singleton accessor, do not init your own instance
+ * Note: Initialization is expected through registerCriteoPublisherId:
+ */
 - (instancetype)init NS_UNAVAILABLE;
+
+/**
+ * Criteo shared instance singleton
+ * Note: Initialization is expected through registerCriteoPublisherId:
+ * @return The Criteo singleton
+ */
 + (nonnull instancetype)sharedCriteo;
 
-/** @abstract Set a custom opt-out/opt-in with same behaviour as the CCPA (US Privacy). */
+/** Set a custom opt-out/opt-in with same behaviour as the CCPA (US Privacy). */
 - (void)setUsPrivacyOptOut:(BOOL)usPrivacyOptOut;
-/** @abstract Set the privacy consent string owned by the Mopub SDK. */
+/** Set the privacy consent string owned by the Mopub SDK. */
 - (void)setMopubConsent:(NSString *)mopubConsent;
 
+/**
+ * Initialize Criteo singleton
+ * @param criteoPublisherId Publisher Identifier
+ * @param adUnits AdUnits array
+ */
 - (void)registerCriteoPublisherId:(NSString *)criteoPublisherId
                       withAdUnits:(NSArray<CRAdUnit *> *)adUnits;
-
+/**
+ * Header bidding API, enrich your request with Criteo metadata
+ * @param request The request to enrich, supports GAM and MoPub
+ * @param adUnit The adUnit related to request
+ */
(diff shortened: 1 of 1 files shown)


CriteoPublisherSdk/Sources/Criteo.h
```diff
@@ -27,20 +27,43 @@
...
+/** Set a custom opt-out/opt-in with same behaviour as the CCPA (US Privacy). */
 - (void)setUsPrivacyOptOut:(BOOL)usPrivacyOptOut;
-/** @abstract Set the privacy consent string owned by the Mopub SDK. */
+/** Set the privacy consent string owned by the Mopub SDK. */
 - (void)setMopubConsent:(NSString *)mopubConsent;
 
+/**
+ * Initialize Criteo singleton
+ * @param criteoPublisherId Publisher Identifier
+ * @param adUnits AdUnits array > TODO deprecate with live bidding?
```
vguerci, line 51: I guess this does not makes sense anymore with live bidding, what about adding a new `registerCriteoPublisherId:` without `adUnits` and mark this one as deprecated?

TurpIF: I fully agree

TurpIF: @Amokrane do you have an opinion on this ?

vguerci: On second thoughts, another approach could be to add instead a `setPublisherId:` like there are other methods such as `setMopubConsent:`. 

By extension I would redeclare these setters into properties (even if the getter is pretty useless) to encourage dot notation use. For instance `[[Criteo sharedCriteo] setUsPrivacyOptOut: NO];` could become `Criteo.sharedCriteo.usPrivacyOptOut = NO;` _(this is not a breaking change as dot notation or setXxxx are already possible)_

TurpIF: In the SDK lifetime some actions are made on `register` (bearcat call, remote config call, prefetch, stuff...)
How would you detect when the SDK registration step is over and when you can start such actions ?

Otherwise, ok for me. I would suggest to strongly indicate to publishers that they should call setTruc and setMachin in the doc of `sharedCriteo`.

vguerci: Ah indeed good point, the current implementation does this implicitly while prefetching.
Let's keep that semantic then, but I would like clarification about what is expected at `register` time and what can wait for first live bid call.

TurpIF: * User agent: it is requested earlier because it is mandatory for bid requests.
* Remote config: as everything is cached, the call is not mandatory (yes we can have a wrong config for the very first call)
* Bearcat: in fact, I just see that at init we're only setting an application lifecycle listener. The listener (starting the bearcat calls) is notified later by the application itself.
* Prefetch: removed soon, but still needed during A/B test
* User consent: publisher should be able to set the user consents at any time: before or after the current register
* Criteo Publisher ID: it is, of course, mandatory for network calls

I think this is all for the registration steps. That's why on Android we hardly split the steps by introducing a builder pattern.

Amokrane: Regarding `prefetch`, just do remember that its deprecation will be effective only when we will confirm the effectiveness of the new strategy through A/B tests (the first A/B test to validate that live-bidding works better than cache only, and the second A/B test later to verify the effect of deleting the prefetch). I have logged a separate ticket on Android side: EE-1251.

Amokrane (changes requested):
Deprecation is not official yet. Needs prod rollout + A/B tests.

CriteoPublisherSdk/Sources/Criteo.h
vguerci: Thanks all for the clarifications.

@TurpIF I took your exhaustive list (thanks!) in consideration while refactoring the code a bit in #108.

Removed my "TODO" about the deprecations, this should be mergeable now.

TurpIF (approved):

Amokrane (approved):

PR #113, opened 2020-10-02
Fix tests not being run by CI

vguerci:
multi_scan does not pick the scheme default testplan as Xcode does.
As a consequence CI was running the Standalone plan introduced
recently, only 78 of 722 tests... ¯\\\_(ツ)_/¯


vguerci:
Yay, I was fearing more broken tests 😎

TurpIF (approved):
image

PR #116, opened 2020-10-06
[Live Bidding] CSM Fix and tests

vguerci:


CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
```diff
@@ -293,7 +293,41 @@ - (void)testFeedbackMessageStateOnEmptyBid {
...
+
+- (void)testFeedbackMessageStateOnLiveBidConsumed {
+  CR_FeedbackMessage *expected = [self.defaultMessage copy];
+  expected.cdbCallEndTimestamp = self.dateInMillisecondsNumber;
+  expected.elapsedTimestamp = self.dateInMillisecondsNumber;
+  expected.cachedBidUsed = YES;
+  expected.zoneId = self.validBid.zoneId;
+
+  self.cdbResponse.cdbBids = @[ self.validBid ];
+  [self prefetchBidWithMockedResponseForAdUnit:self.adUnit];
```
TurpIF, line 306: I think you don't need to prefetch for live bidding

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
```diff
@@ -388,6 +422,89 @@ - (void)testReadyToSendOnBidStateOnEmptyBid {
...
+
+- (void)testReadyToSendOnLiveBidConsumed {
+  self.cdbResponse.cdbBids = @[ self.validBid ];
+  [self configureApiHandlerMockWithCdbRequest:self.cdbRequest
+                                  cdbResponse:self.cdbResponse
+                                        error:nil];
+  [self fetchLiveBidForAdUnit:self.adUnit];
+
+  XCTAssertEqual(self.feedbackFileManagingMock.readWriteDictionary.count, 0);
+  XCTAssertEqual([self.lastSentMessages count], 1);
```
TurpIF, line 494: Nit: self.lastSendMessages.count

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
```diff
@@ -388,6 +422,89 @@ - (void)testReadyToSendOnBidStateOnEmptyBid {
...
+
+- (void)testReadyToSendOnLiveBidStateOnEmptyBid {
+  self.cdbResponse.cdbBids = @[];
+  [self configureApiHandlerMockWithCdbRequest:self.cdbRequest
+                                  cdbResponse:self.cdbResponse
+                                        error:nil];
+  [self fetchLiveBidForAdUnit:self.adUnit];
+
+  XCTAssertEqual(self.feedbackFileManagingMock.readWriteDictionary.count, 0);
+  XCTAssertEqual([self.lastSentMessages count], 1);
```
TurpIF, line 505: Nit: self.lastSendMessages.count

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
```diff
@@ -388,6 +422,89 @@ - (void)testReadyToSendOnBidStateOnEmptyBid {
...
+  [self configureApiHandlerMockWithCdbRequest:self.cdbRequestForInvalidBid
+                                  cdbResponse:self.cdbResponseWithInvalidBid
+                                        error:nil];
+  [self fetchLiveBidForAdUnit:self.adUnit];
+
+  XCTAssertEqual(self.feedbackFileManagingMock.readWriteDictionary.count, 0);
+  XCTAssertEqual(self.lastSentMessages.count, 1);
+}
+
+- (void)testReadyToSendOnLiveBidConsumed {
```
TurpIF, line 486: Isn't this a duplicate of `testReadyToSendOnLiveBidValidBidReceived` ?
For CSM, this sentence has more sense: on bid consumed

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
```diff
@@ -388,6 +422,89 @@ - (void)testReadyToSendOnBidStateOnEmptyBid {
...
+                                  cdbResponse:self.cdbResponse
+                                        error:nil];
+
+  [self fetchLiveBidForAdUnit:self.adUnit];
+
+  XCTAssertEqual(self.feedbackFileManagingMock.readWriteDictionary.count, 0);
+  XCTAssertEqual(self.lastSentMessages.count, 1);
+}
+
+- (void)testReadyToSendOnLiveBidNoBidReceived {
```
TurpIF, line 439: Isn't this a dup of `testReadyToSendOnLiveBidStateOnEmptyBid` ?
I think that this sentence make more sense: no bid over empty bid

TurpIF (changes requested):

CriteoPublisherSdk/Tests/IntegrationTests/CR_FeedbackMessageFunctionalTests.m
```diff
@@ -281,7 +327,10 @@ - (void)getBidAndWaitForFetchAndFeedbackWithAdUnit:(CRAdUnit *)adUnit {
 
   CR_CacheAdUnit *cacheAdUnit = [CR_AdUnitHelper cacheAdUnitForAdUnit:adUnit];
   [self.criteo getBid:cacheAdUnit
-      responseHandler:^(CR_CdbBid *bid){
+      responseHandler:^(CR_CdbBid *bid) {
+        if (self.config.isLiveBiddingEnabled) {
+          [self.dependencyProvider.feedbackDelegate sendFeedbackBatch];
```
TurpIF, line 366: Shouldn't this be already done by the bid manager ?

TurpIF (changes requested):

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
```diff
@@ -230,12 +233,14 @@ - (void)testFeedbackMessageStateOnInvalidBid {
                                   cdbResponse:self.cdbResponseWithInvalidBid
                                         error:nil];
 
-  [self.bidManager prefetchBidForAdUnit:self.adUnitForInvalidBid];
+  [self fetchBidForAdUnit:self.adUnit];
```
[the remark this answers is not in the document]
vguerci: indeed, thanks! 😅

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
```diff
@@ -428,4 +433,14 @@ - (void)invokeBeforeCdbHandlerOnBidRequestWithCdbRequest:(CR_CdbRequest *_Nullab
                      completionHandler:[OCMArg any]]);
 }
 
+- (void)fetchBidForAdUnit:(CR_CacheAdUnit *)adUnit {
+  [self fetchBidsForAdUnits:@[ adUnit ]];
```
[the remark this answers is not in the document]
vguerci: Removed

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
vguerci: This method is used to enqueue a metric that will be sent with next live bid request.
I agree prefetching isn't the most logical thing in these tests, but the goal is in fact to prepare CSM queue state.
Also, that comes from legacy tests, which has the comment on `prefetchBidWithMockedResponseForAdUnit`:
```// TODO: improve tests relying on this so that we don't need to call prefetchBid.```

File: CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m (hunk 2 above)
Comment (vguerci) [reply CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m:494]: Done

File: CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m (hunk 3 above)
Comment (vguerci) [reply CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m:505]: Done

File: CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m (hunk 4 above)
Comment (vguerci) [reply CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m:486]: Done

File: CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m (hunk 5 above)
Comment (vguerci) [reply CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m:439]: Done

File: CriteoPublisherSdk/Tests/IntegrationTests/CR_FeedbackMessageFunctionalTests.m (hunk 6 above)
Comment (vguerci) [reply CriteoPublisherSdk/Tests/IntegrationTests/CR_FeedbackMessageFunctionalTests.m:366]: Nope, as I've kept the same logic as before, to send metrics with live request, not after, meaning we still have a delay with metrics.
I'd like to change, but a quick attempt showed that wasn't trivial mostly because of asynchronicity, but also because the fetching logic is shared across prefetch and live bid.
What's your take about it? How is this handled on Android side?

File: CriteoPublisherSdk/Tests/IntegrationTests/CR_FeedbackMessageFunctionalTests.m [hunk 9]
```diff
@@ -217,6 +219,9 @@ - (void)testGivenFeedbackRequestError_whenGettingBids_thenFeedbackRequest {
   [self getBidAndWaitForFetchAndFeedbackWithAdUnit:adUnit2];
   [self getBidAndWaitForFetchAndFeedbackWithAdUnit:adUnit2];
 
+  if (self.config.isLiveBiddingEnabled) {
+    [self sendFeedbacksAndWaitFeedbackSent];
```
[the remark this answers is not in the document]
vguerci: See previous comment

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerFeedbackTests.m
TurpIF: Ok

CriteoPublisherSdk/Tests/IntegrationTests/CR_FeedbackMessageFunctionalTests.m
TurpIF: Ok, understood. On Android CSM are sent in parallel of bid requests. So it depends of this concurrent race :D
In the test we tried to have a deterministic approach by doing N+1 bids to be sure to have N metrics. But it makes test harder to read and to maintain finally.
So up to you. Maybe could you just keep this and add a small comment explaining the why ?

CriteoPublisherSdk/Tests/IntegrationTests/CR_FeedbackMessageFunctionalTests.m
```diff
@@ -243,16 +248,13 @@ - (void)testGivenFeedbackRequestError_whenGettingBids_thenFeedbackRequest {
     return value[@"requestGroupId"];
   }];
   AssertArrayWithUniqueElements(impressionIds, 3);
-  if (self.config.isLiveBiddingEnabled) {
```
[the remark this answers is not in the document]
TurpIF: When time budget is exceeded, when bids are put in the cache but the `requestGroupIds` field in metrics should be equal to `id` in the CDB request. So even if bids are cached, with live bidding, they are always coming from different requests, no ?

Maybe could we tchat about this.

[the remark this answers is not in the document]
vguerci: Yes but not in this scenario with the initial prefetch with two adUnits. Reason why it ends up in the same state as cache bidding.

CriteoPublisherSdk/Tests/IntegrationTests/CR_FeedbackMessageFunctionalTests.m
vguerci: So to clarify same thing on iOS with bid and csm are sent in //.

Ok for a comment, I might try to sequence bid then csm FAR as it is the most logical thing to do in live bid mode.
If not, maybe fill an issue for the future?

TurpIF (approved):

vguerci:
Thanks again! 🚀

PR #117, opened 2020-10-06
Update default value of time budget

TurpIF:
JIRA: EE-1282


CriteoPublisherSdk/Sources/CRConstants.h
```diff
@@ -23,6 +23,6 @@
 #define CRITEO_PUBLISHER_SDK_VERSION @"4.0.0-alpha2"
 
 #define CRITEO_DEFAULT_BID_TTL_IN_SECONDS 15 * 60
-#define CRITEO_DEFAULT_LIVE_BID_TIME_BUDGET_IN_SECONDS 1.0f
+#define CRITEO_DEFAULT_LIVE_BID_TIME_BUDGET_IN_SECONDS 8.0f
```
Amokrane, line 26: nit: we should document why 8 secs was picked as the default. A link to the JIRA where the ticket was held is also fine.

TurpIF: The JIRA is already in the commit message. Isn't this sufficient ?
Moreover, the analysis that was done is only valid for now. Later, the results may change a little bit.

vguerci (approved):

PR #118, opened 2020-10-07
[Live Bidding] Rework token logic into consumable bids

vguerci:


CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerTests.m
```diff
@@ -216,7 +216,6 @@ - (void)testBidResponseForEmptyBid {
 
   XCTAssertEqualWithAccuracy(bid.price, 0.0f, 0.1);
   XCTAssertNil(bid.bidToken);
```
TurpIF: Don't you need to remove this bid.bidToken too ?

TurpIF (approved):
You may simplify a little bit the logic by having only the AdUnitType in the Bid instead of the AdUnit.

vguerci:
> You may simplify a little bit the logic by having only the AdUnitType in the Bid instead of the AdUnit.

I'm not sure that is possible as If I remember well we also use it on InHouse `CRBannerView.loadWithBid:` to check that it `isEqual` to the one set on view, which requires for instance size.

Thanks for your review :)

PR #119, opened 2020-10-07
[Live Bidding] App Bidding integration

vguerci:
JIRA: EE-1264

### DIFF
--- a/CriteoPublisherSdk/Sources/CR_BidManager.m
+++ b/CriteoPublisherSdk/Sources/CR_BidManager.m
@@ -305,27 +305,6 @@ typedef void (^CR_CdbResponseHandler)(CR_CdbResponse *response);
   }
 }
 
-- (void)addCriteoBidToRequest:(id)adRequest forAdUnit:(CR_CacheAdUnit *)adUnit {
-  @try {
-    [self unsafeAddCriteoBidToRequest:adRequest forAdUnit:adUnit];
-  } @catch (NSException *exception) {
-    CLogException(exception);
-  }
-}
-
-- (void)unsafeAddCriteoBidToRequest:(id)adRequest forAdUnit:(CR_CacheAdUnit *)adUnit {
-  if (!config) {
-    CLog(@"Config hasn't been fetched. So no bids will be fetched.");
-    return;
-  } else if ([config killSwitch]) {
-    CLog(@"killSwitch is engaged. No bid will be fetched.");
-    return;
-  }
-
-  CR_CdbBid *fetchedBid = [self getBidThenFetch:adUnit];
-  [self.headerBidding enrichRequest:adRequest withBid:fetchedBid adUnit:adUnit];
-}
-
 - (CRBid *)bidForAdUnit:(CRAdUnit *)adUnit {
   CR_CacheAdUnit *cacheAdUnit = [CR_AdUnitHelper cacheAdUnitForAdUnit:adUnit];
   CR_CdbBid *bid = [self getBidThenFetch:cacheAdUnit];
--- a/CriteoPublisherSdk/Sources/CR_BidManager.m
+++ b/CriteoPublisherSdk/Sources/CR_BidManager.m
@@ -17,12 +17,13 @@
 // limitations under the License.
 //
 
+#import "CRBid+Internal.h"
+#import "CR_AdUnitHelper.h"
 #import "CR_BidManager.h"
-#import "Logging.h"
 #import "CR_FeedbackController.h"
 #import "CR_HeaderBidding.h"
 #import "CR_ThreadManager.h"
-#import "CR_AdUnitHelper.h"
+#import "Logging.h"
 
 typedef void (^CR_CdbResponseHandler)(CR_CdbResponse *response);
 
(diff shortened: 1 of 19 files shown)


CriteoPublisherSdk/Sources/CR_BidManager.m
```diff
@@ -288,6 +289,30 @@ - (void)updateTimeToNextCallIfProvided:(CR_CdbResponse *)cdbResponse {
...
+  @try {
+    [self enrichUnsafelyAdObject:object withBid:bid];
+  } @catch (NSException *exception) {
+    CLogException(exception);
+  }
+}
+
+- (void)enrichUnsafelyAdObject:(id)object withBid:(CRBid *)bid {
+  if (!config) {
+    CLog(@"Config hasn't been fetched. So no bids will be fetched.");
```
TurpIF, line 302: Bid is already fetched. I guess that such conditions are not valid anymore with the new logic. Same thing for the kill switch.

Moreover, I thought that bidding should not have to wait for the config to be fetched.

CriteoPublisherSdk/Tests/IntegrationTests/CR_IntegrationsTestBase.m
```diff
@@ -39,9 +47,19 @@ - (void)waitForIdleState {
...
 
-- (void)tearDown {
-  [self waitForIdleState];
-  [super tearDown];
+#pragma mark - App Bidding helper
+
+- (void)enrichAdObject:(id)object
+             forAdUnit:(CRAdUnit *)adUnit
+           thenWaitFor:(void (^)(void))completionBlock {
+  [self cr_waitForExpectationBlock:^(XCTestExpectation *expectation) {
```
TurpIF, line 55: You're doing all your assertions inside a block which make the tests a little bid harder to read (when the block is too big).
What do you think about mocking a synchronous call ? Either via XCTestExpectation or through `[self waitForIdleState]`.

CriteoPublisherSdk/Tests/IntegrationTests/CR_ProfileIdFunctionalTests.m
```diff
@@ -312,4 +327,17 @@ - (NSDictionary *)csmRequest {
   }];
 }
 
+- (void)enrichAd:(id)adObject
+       forAdUnit:(CRAdUnit *)adUnit
+     thenWaitFor:(void (^)(void))completionBlock {
+  [self cr_waitForExpectationBlock:^(XCTestExpectation *expectation) {
```
TurpIF, line 333: Same comment than in CR_IntegrationTestBase. What do you think about mocking a synchronous call to make test easier to read ?

I see that test are built with `waitForIdleState`. Then tests should already be sync, no ?

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerTests.m
```diff
@@ -406,7 +407,7 @@ - (void)testAddCriteoBidToRequestCallHeaderBidding {
 - (void)testAddCriteoBidToRequestWhenKillSwitchIsEngagedShouldNotEnrichRequest {
```
TurpIF, line 407: I'm not sure that this scenario should be kept. If killSwitch is on, you won't even get a bid. Maybe should you either remove the test, either improve it to test getBid+enrichAdObject.

TurpIF (changes requested):

CriteoPublisherSdk/Sources/CR_BidManager.m
vguerci: > Bid is already fetched. I guess that such conditions are not valid anymore with the new logic. Same thing for the kill switch.

For some reason I was thinking the killswitch wasn't meant only for bid requests. Will change this.

> Moreover, I thought that bidding should not have to wait for the config to be fetched.

Not sure how this relates to this particular code as there is no wait there?
Are you talking about the bid fetching general case?

CriteoPublisherSdk/Tests/IntegrationTests/CR_IntegrationsTestBase.m
vguerci: This is mostly a way to adapt all tests w/o being too intrusive. While it is maybe not the best one, that ease reviewing, and works with all scenarios, including I think ones that wasn't relying on idleState. I'll see if I can change this w/o too much rewrite / time investment.

CriteoPublisherSdk/Tests/IntegrationTests/CR_ProfileIdFunctionalTests.m
vguerci: Same as previous comment, not all tests are based on idle state, this approach works for all scenarios.

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerTests.m
vguerci: Will remove with the logic tested.

CriteoPublisherSdk/Sources/CR_BidManager.m
TurpIF: > > Bid is already fetched. I guess that such conditions are not valid anymore with the new logic. Same thing for the kill switch.
> 
> For some reason I was thinking the killswitch wasn't meant only for bid requests. Will change this.

You're right, the kill switch is for deactivating the whole SDK. But IMHO just cutting the network+bidding logic (so `bid=nil`) is enough to deactivate the SDK.
So ok for keeping or removing the killSwitch check.

> 
> > Moreover, I thought that bidding should not have to wait for the config to be fetched.
> 
> Not sure how this relates to this particular code as there is no wait there?
> Are you talking about the bid fetching general case?

I was talking about the `if (!config)`. This condition should not belong to this method as you already have a bid. But I'm even questioning if it belongs to the `loadBid` method: bidding+config are async, so bidding should not care if config was fetched or not.

vguerci: Thanks for the details, will remove then as you said it is covered anyway on bid requesting side which is more central and logical.

CriteoPublisherSdk/Tests/IntegrationTests/CR_IntegrationsTestBase.m
vguerci: So your suggestion works in the end, took a bit more time that I was hoping, but anyway, that's better now, thank you :)

CriteoPublisherSdk/Tests/UnitTests/CR_BidManagerTests.m
vguerci: Done

CriteoPublisherSdk/Tests/IntegrationTests/CR_ProfileIdFunctionalTests.m
vguerci: Done

CriteoPublisherSdk/Sources/CR_BidManager.m
vguerci: Done

TurpIF (approved):

PR #121, opened 2020-10-09
[Live Bidding] Faster tests

vguerci:


TurpIF (dismissed):
LGTM

Although, I'm not sure if this scale correctly. What do you think about doing the inverse ? Having a shot time budget by default and explicitly ask for a longer one if needed.

So when adding a test that don't care, then time budget is low and won't slow the test.
When adding a test that want a normal time budget to not timeout, then the test fail notifying that explicit config should be made.

vguerci:
> Although, I'm not sure if this scale correctly. What do you think about doing the inverse ? Having a shot time budget by default and explicitly ask for a longer one if needed.

Agreed, but I envisaged several approaches to this fix, and this is the only one not intrusive to production code path I could figure, maybe you have another idea?
Also this is partially a default as it is on this [`Criteo+Testing` code path](https://github.com/criteo/ios-publisher-sdk/blob/5d86424e4d00cb5db3ba11f4d9c95ae8a9ce5feb/CriteoPublisherSdk/Tests/Utility/Categories/Criteo%2BTesting.m#L88).

TurpIF:
> > Although, I'm not sure if this scale correctly. What do you think about doing the inverse ? Having a shot time budget by default and explicitly ask for a longer one if needed.
> 
> Agreed, but I envisaged several approaches to this fix, and this is the only one not intrusive to production code path I could figure, maybe you have another idea?
> Also this is partially a default as it is on this [`Criteo+Testing` code path](https://github.com/criteo/ios-publisher-sdk/blob/5d86424e4d00cb5db3ba11f4d9c95ae8a9ce5feb/CriteoPublisherSdk/Tests/Utility/Categories/Criteo%2BTesting.m#L88).

Why not decorating [`CR_DependencyProvider#testing_dependencyProvider`](https://github.com/criteo/ios-publisher-sdk/blob/5d86424e4d00cb5db3ba11f4d9c95ae8a9ce5feb/CriteoPublisherSdk/Tests/Utility/Categories/CR_DependencyProvider%2BTesting.m#L31) instead of `Criteo+Testing` ?

The `testing_dependencyProvider` could set a small timebudget by default. Or is it too global ?

vguerci:
> The `testing_dependencyProvider` could set a small timebudget by default. Or is it too global ?

Good idea, let's try 🧨

vguerci:
And it works! 🎉 🤘 🚀

TurpIF (approved):

PR #132, opened 2020-10-27
[Skan] Read adNetworkIds

vguerci:


CriteoPublisherSdk/Sources/Configuration/SkanInfo.swift
```diff
@@ -0,0 +1,44 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+struct SkanInfo {
```
TurpIF, line 22: Out of context, Skan is not expressive and hard to understand (like the ttnc stuff you removed earlier).
What do you think about renaming this to a longer form ?

CriteoPublisherSdk/Sources/Criteo.swift
```diff
@@ -0,0 +1,36 @@
...
+// limitations under the License.
+//
+
+import Foundation
+
+@objc extension Criteo {
+  public func validateIntegration() {
+    let skanInfo = SkanInfo()
+    if !skanInfo.hasCriteoId {
+      print(
```
TurpIF, line 26: I remember that there is a logger macro for warnings in the project. Can't you use it ?

TurpIF (changes requested):
LGTM except for the naming of `Skan`

CriteoPublisherSdk/Sources/Criteo.swift
vguerci: Macros does not exist in swift ;p

CriteoPublisherSdk/Sources/Configuration/SkanInfo.swift
vguerci: We already use the `Skan` acronym everywhere, (unlike the ttnc).

Also I was reluctant to name this following the Apple naming prefix `SKAdNetwork` as it could be confusing or worse name clashing. But since this will be in a different module, that shouldn't be an issue.
Let's even push further that and try a `SKAdNetwork` extension 😂.

CriteoPublisherSdk/Sources/Criteo.swift
vguerci: This will be reworked anyway while we address the Logging topic. Which current Macros aren't fitting.

CriteoPublisherSdk/Sources/Configuration/SkanInfo.swift
vguerci: Done in this [push](https://github.com/criteo/ios-publisher-sdk/compare/6b1aaa4b6601c57cb146406b0ac05b16334c3466..73d60fa392d6fcf899c732d3d280dc5122db7e7c)

TurpIF (approved):

PR #135, opened 2020-10-30
[Actions] Factorize with local composite actions

vguerci:


vguerci:
Not yet possible, `uses` isn't supported yet inside a composite action

PR #137, opened 2020-11-02
[InHouse] Allow init w/o adUnit

vguerci:
As it is redundant with loadWithBid. As a consequence, remove check
that adUnit is the one that was used at init

JIRA: EE-1332


CriteoPublisherSdk/Sources/Native/CRNativeLoader.m
```diff
@@ -41,6 +41,13 @@ @implementation CRNativeLoader
 
 #pragma mark - Life cycle
 
+- (instancetype)init {
```
TurpIF, line 44: Using this init does not initialize `self.criteo` or `self.urlOpener`. Is it intended ?
Generally when overloading constructors, constructor will less parameters calls the one with more parameters with default value or nil (nil here). So the object is always initialized in the same manner.

TurpIF (changes requested):
What do you think about updating the InHouse functional tests to use the constructors without ad unit ?
So this constructor would be under test.

CriteoPublisherSdk/Sources/Native/CRNativeLoader.m
vguerci: Done.

vguerci:
> What do you think about updating the InHouse functional tests to use the constructors without ad unit ?
> So this constructor would be under test.

This functional test use a shared utility checker that does the init, so it wasn't really possible. Instead added some unit tests on the class test suites which effectively tests `loadAdWithBid:` w/o an adUnit.

vguerci:
Also added some checks to prevent `loadAd` w/o and adUnit on standalone.

TurpIF (approved):

vguerci:
Thanks!

PR #140, opened 2020-11-05
Implement public API collecting ContextData from publisher

TurpIF:
JIRA: EE-1323

### DIFF
--- a/CriteoAdViewer/Sources/AdViewer/AdConfig.swift
+++ b/CriteoAdViewer/Sources/AdViewer/AdConfig.swift
@@ -41,3 +41,10 @@ struct AdConfig {
     }
   }
 }
+
+func defaultContextData() -> CRContextData {
+  CRContextData(dictionary: [
+    CRContextDataContentUrl: "http://foo.bar",
+    "data.dummy": 42
+  ])
+}
--- a/CriteoPublisherSdk/Sources/CR_BidManager.h
+++ b/CriteoPublisherSdk/Sources/CR_BidManager.h
@@ -35,6 +35,7 @@
 #import "CRBid+Internal.h"
 #import "CR_FeedbackStorage.h"
 #import "Criteo+Internal.h"
+#import "CRContextData.h"
 
 @protocol CR_FeedbackDelegate;
 @class CR_HeaderBidding;
@@ -75,23 +76,26 @@
  * - Returned synchronously using the cache bidding strategy (prefetch / get from cache)
  *
  * @param adUnit The ad unit to request a bid
+ * @param contextData The context for this bid
  * @param responseHandler The block called once bid has been loaded
  * Note: responseHandler is potentially not invoked on main queue
  */
 - (void)loadCdbBidForAdUnit:(CR_CacheAdUnit *)adUnit
+                    context:(CRContextData *)contextData
             responseHandler:(CR_CdbBidResponseHandler)responseHandler;
 
 @end
 
 @interface CR_BidManager (CacheBidding)
 
-- (void)prefetchBidsForAdUnits:(CR_CacheAdUnitArray *)adUnits;
+- (void)prefetchBidsForAdUnits:(CR_CacheAdUnitArray *)adUnits context:(CRContextData *)contextData;
 
 @end
 
 @interface CR_BidManager (LiveBidding)
 
 - (void)fetchLiveBidForAdUnit:(CR_CacheAdUnit *)adUnit
+                      context:(CRContextData *)contextData
               responseHandler:(CR_CdbBidResponseHandler)responseHandler;
 
 @end
--- a/CriteoPublisherSdk/Sources/CR_BidManager.h
+++ b/CriteoPublisherSdk/Sources/CR_BidManager.h
@@ -81,21 +81,22 @@
  * Note: responseHandler is potentially not invoked on main queue
  */
 - (void)loadCdbBidForAdUnit:(CR_CacheAdUnit *)adUnit
-                    context:(CRContextData *)contextData
+                withContext:(CRContextData *)contextData
             responseHandler:(CR_CdbBidResponseHandler)responseHandler;
 
 @end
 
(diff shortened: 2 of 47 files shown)


CriteoPublisherSdk/Sources/Context/CRContextData.m
```diff
@@ -19,6 +19,33 @@
...
+
 @implementation CRContextData
 
-@end
\ No newline at end of file
++ (NSString *)contentUrl {
+  return @"content.url";
+}
+
++ (CRContextData *)dataWithDictionary:(NSDictionary<NSString *, id> *)dictionary {
```
vguerci, line 36: NIT: Naming, `contextDataWithDictionary:`

vguerci (changes requested):
Nicely done! Not much to say, just the naming that is a bit inconsistent I think, `context:`, `contextData:` and `loadAdWithContext:`. I think it would be preferable to use `withContext:` everywhere.

That leads me to think that the context class could have been simply named `CRContext`, `Data` not bringing much value, but that is clearly NIT / not a requested change.

Oh also, I have seen some (temporary) lines of Swift code with nil arguments, I would suggest to verify that this is properly checked at compile time. (Can do this if you want)

TurpIF:
> Nicely done! Not much to say, just the naming that is a bit inconsistent I think, `context:`, `contextData:` and `loadAdWithContext:`. I think it would be preferable to use `withContext:` everywhere.
> 
> That leads me to think that the context class could have been simply named `CRContext`, `Data` not bringing much value, but that is clearly NIT / not a requested change.

Agreed to use `withContext` everywhere it is simpler. For `CRContext`, this would force us to have `CRUser` instead of `CRUserData`. Don't you this that `CRUser` is kind of too general ?
 
> Oh also, I have seen some (temporary) lines of Swift code with nil arguments, I would suggest to verify that this is properly checked at compile time. (Can do this if you want)

?? Are you talking about the `context: nil /* TODO */` I've put in intermediate commits ? Yes it makes the compilation fail. So intermediate commits are not passing the CI. I've done this way to ease the dev and review process. If it is an issue, I can squash the commits.

CriteoPublisherSdk/Sources/Context/CRContextData.m
TurpIF: Done

vguerci:
> Agreed to use `withContext` everywhere it is simpler. For `CRContext`, this would force us to have `CRUser` instead of `CRUserData`. Don't you this that `CRUser` is kind of too general ?

`CRUserContext` could be more adapted, as the context is the `Data`. But that is really NIT, I'm fine with `ContextData`, especially if you have the same on Android side.

> ?? Are you talking about the `context: nil /* TODO */` I've put in intermediate commits ? Yes it makes the compilation fail. So intermediate commits are not passing the CI. I've done this way to ease the dev and review process. If it is an issue, I can squash the commits.
Yup I was talking about this, looking at commits I thought it was building, It is ok if it does not. No need to squash.

TurpIF:
> > Agreed to use `withContext` everywhere it is simpler. For `CRContext`, this would force us to have `CRUser` instead of `CRUserData`. Don't you this that `CRUser` is kind of too general ?
> 
> `CRUserContext` could be more adapted, as the context is the `Data`. But that is really NIT, I'm fine with `ContextData`, especially if you have the same on Android side.

Ok, I keep them to be like on Android.

I also added the `initWithDico` and migrate HTML -> obj-c doc.

CriteoPublisherSdk/Sources/CR_BidManager.h
```diff
@@ -75,23 +76,26 @@
  * - Returned synchronously using the cache bidding strategy (prefetch / get from cache)
  *
  * @param adUnit The ad unit to request a bid
+ * @param contextData The context for this bid
  * @param responseHandler The block called once bid has been loaded
  * Note: responseHandler is potentially not invoked on main queue
  */
 - (void)loadCdbBidForAdUnit:(CR_CacheAdUnit *)adUnit
+                    context:(CRContextData *)contextData
```
vguerci, line 84: Sorry If I wasn't clear, but I meant `withContext:` for these kind of signatures as well (just add a commit don't bother amending each)

CriteoPublisherSdk/Sources/Context/CRContextData.m
```diff
@@ -0,0 +1,49 @@
...
+
+@interface CRContextData ()
+
+@property(strong, nonatomic, readonly) NSDictionary<NSString *, id> *data;
+
+@end
+
+@implementation CRContextData
+
++ (NSString *)contentUrl {
```
vguerci, line 30: I am sorry, noticing this a bit late, but I thought at first it was a typed property on Context, i.e. `self.contextData.contentUrl = @"http://whatev.er"`.
The naming is confusing. Usually such constants are declared as top level one, like [this](https://developer.apple.com/documentation/storekit/skstoreproductparameteradnetworkidentifier?language=objc). In this case this should be something like: `NSString *const CRContextDataContentUrl`

vguerci (changes requested):

CriteoPublisherSdk/Sources/CR_BidManager.h
TurpIF: Done

CriteoPublisherSdk/Sources/Context/CRContextData.m
TurpIF: Done

CriteoAdViewer/Sources/AdViewer/AdConfig.swift
```diff
@@ -41,3 +41,10 @@ struct AdConfig {
     }
   }
 }
+
+func defaultContextData() -> CRContextData {
```
vguerci, line 45: NIT: this probably could have been expressed as a `let defaultContextData`

vguerci (approved):

PR #143, opened 2020-11-09
[SkAdNetwork] URL resolver

vguerci:


CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
```diff
@@ -0,0 +1,121 @@
...
+#import "CR_DeviceInfo.h"
+#import "CR_URLRequest.h"
+
+@interface CR_URLResolution ()
+@property(nonatomic, strong) NSURL *URL;
++ (instancetype)resolutionWithType:(CR_URLResolutionType)type;
++ (instancetype)resolutionError:(NSError *)error;
+@end
+
+@interface CR_URLResolver ()
```
TurpIF, line 31: Is this useful ? Nothing is declared in this category

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
```diff
@@ -0,0 +1,121 @@
...
+@end
+
+@implementation CR_URLResolver
+
+#pragma mark - Lifecycle
+
+- (instancetype)init {
+  self = [super init];
+  if (self) {
+    _urlSession =
```
TurpIF, line 46: Shouldn't you share this session with the one used in the NetworkManager class ?
You might declare the session in the dependency provider and provide it to both NetworkManager and URLResolver.

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
```diff
@@ -0,0 +1,121 @@
...
+  [resolver resolverURL:url deviceInfo:deviceInfo resolution:resolution];
+}
+
+- (void)resolverURL:(NSURL *)url
+         deviceInfo:(CR_DeviceInfo *)deviceInfo
+         resolution:(CR_URLResolutionHandler)resolution {
+  self.resolution = resolution;
+  CR_URLRequest *request = [CR_URLRequest requestWithURL:url deviceInfo:deviceInfo];
+  NSURLSessionDataTask *task = [self.urlSession dataTaskWithRequest:request];
+  [task resume];
```
TurpIF, line 69: You can add a completion handler in the task and use it with the threadManager. So the threadManager can be used to wait on such task.

See https://github.com/criteo/ios-publisher-sdk/blob/main/CriteoPublisherSdk/Sources/Network/CR_NetworkManager.m#L85

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
```diff
@@ -0,0 +1,121 @@
...
+  completionHandler([self shouldRedirectTo:request] ? request : nil);
+}
+
+- (void)URLSession:(NSURLSession *)session
+                    task:(NSURLSessionTask *)task
+    didCompleteWithError:(nullable NSError *)error {
+  if (error != nil) {
+    self.resolution([CR_URLResolution resolutionError:error]);
+    return;
+  }
```
TurpIF, line 88: Nit: a if/else can be more readable, don't you think ?

TurpIF (changes requested):
Do you know if we should handle infinite loop ? We can handle it very naively: after 42 redirects, we drop the connection (I hop we're not doing 32 or 64 redirections to encode an integer :D)

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: Useless, will be removed

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: Ideally yes, session should be shared to leverage cache. This is a deliberate choice and tradeoff for simplification because a session has only one delegate for all the tasks, it become rapidly complex. And on the other hand, from my understanding, when reusing the session, in the url resolver context, the benefits are small.

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: The completion handler is only called when the request is fully loaded, which includes redirections. The main purpose of the resolver is the [`willPerformHTTPRedirection:`](https://developer.apple.com/documentation/foundation/nsurlsessiontaskdelegate/1411626-urlsession) delegate message _before_ the url is loaded, I hardly see how it can be done the way you suggest, but maybe you have another idea?

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: I think this is more a matter of personal preference, I'm following our code style, ["Golden path"](https://github.com/raywenderlich/objective-c-style-guide#golden-path).

vguerci: > Do you know if we should handle infinite loop ? We can handle it very naively: after 42 redirects, we drop the connection (I hop we're not doing 32 or 64 redirections to encode an integer :D)

Not needed as it is already handled by iOS. Added a test that ensure this (observed max redirects is 20).

vguerci:
> Do you know if we should handle infinite loop ? We can handle it very naively: after 42 redirects, we drop the connection (I hop we're not doing 32 or 64 redirections to encode an integer :D)

Not needed as it is already handled by iOS. Added a test that ensure this (observed max redirects is 20).

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
TurpIF: Ok then

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
TurpIF: Not an issue, the idea would be to have a way to wait for the end of the processus (so when redirections are over) so that tests can observe an idle state (simpler to assert than during a dynamic state).

Although, this is only useful if we have a functional scenario such as:
- start a stuff
  - stuff trigger redirection
  - once redirection is finished, post a foo
- wait for idle state
- observe that foo is posted

I don't know if you plan to write such a test, if not then ok to skip this.

TurpIF (approved):
LGTM, I let you chose to add or not the threadManager if you need it for further tests.

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: I plan to mock the resolver for tests that will depend on it.

To be honest, generally I prefer to use the "wait for `XCTTestExpectation`" approach that is the "standard" way to test asynchronous code on iOS platform. Most of the case I find it simpler and less intrusive than the threadmanager. Thread manager makes more sense when queuing / concurrency is involved.

vguerci:
> LGTM, I let you chose to add or not the threadManager if you need it for further tests.

As said, not sure it brings values yet. But will for sure in the future commits if it does.
Thank you :)

PR #144, opened 2020-11-10
Implement public API collecting UserData from publisher

TurpIF:


CriteoPublisherSdk/Sources/Context/CRUserData.m
```diff
@@ -0,0 +1,47 @@
...
+
+@property(strong, nonatomic, readonly) NSDictionary<NSString *, id> *data;
+
+- (instancetype)initWithDictionary:(NSDictionary<NSString *, id> *)dictionary;
+
+@end
+
+@implementation CRUserData
+
++ (CRUserData *)dataWithDictionary:(NSDictionary<NSString *, id> *)dictionary {
```
vguerci, line 41: NIT: Naming `userDataWithDictionary:`

CriteoPublisherSdk/Sources/Public/CRUserData.h
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CRUserData : NSObject
+
++ (CRUserData *)dataWithDictionary:(NSDictionary<NSString *, id> *)dictionary;
```
vguerci, line 76: I would suggest to expose the counterpart instance method `initWithDictionary` as well, as it is often what the developers are looking for and also because this is what I see in the style of most of the SDK.
I realise this applies as well to the `CRContextData` class.

CriteoPublisherSdk/Tests/UnitTests/CriteoTests.m
```diff
@@ -28,11 +28,14 @@
...
 #import "CR_AppEvents.h"
 #import "CR_BidManager.h"
+#import "CR_UserDataHolder.h"
+#import "CRUserData+Internal.h"
 
 @interface CriteoTests : XCTestCase
 
 @property(strong, nonatomic) Criteo *criteo;
 @property(strong, nonatomic) CR_IntegrationRegistry *integrationRegistry;
+@property(strong, nonatomic) CR_UserDataHolder *userDataHolder;
```
vguerci, line 38: What do you think about marking this property as `atomic` and directly use the `CRUserData` here and remove the `CR_UserDataHolder` class?

CriteoPublisherSdk/Sources/Public/CREmailHasher.h
```diff
@@ -0,0 +1,42 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CREmailHasher : NSObject
+
+/**
+ * Helper function to hash emails for `CRUserData.hashedEmail`
+ *
+ * <h1>Hashing Format</h1>
+ * The hashing should be the users’ email address:
+ * <ul>
```
vguerci, line 31: NIT: I don't think html is supported in obj-C documention

vguerci (changes requested):

CriteoPublisherSdk/Sources/Public/CREmailHasher.h
TurpIF: Done

CriteoPublisherSdk/Sources/Public/CRUserData.h
TurpIF: Done

CriteoPublisherSdk/Sources/Context/CRUserData.m
TurpIF: Done

CriteoPublisherSdk/Tests/UnitTests/CriteoTests.m
TurpIF: Do you mean that the dependency provider should hold the value ?
But then the ApiHandler should have a direct dependency to the DependencyProvider.

vguerci: Yes sorry this comment should have gone on the dependency provider.
Let's keep it that way then, unless maybe there isn't a strong requirement not to depend on the provider directly?

vguerci (dismissed):

TurpIF:
I rebased the branch on top of `main`. Please, can I get again your review @vguerci ?

CriteoPublisherSdk/Tests/UnitTests/CriteoTests.m
TurpIF: On Android, we definitely do not want to depend directly on the provider. In this way, we can directly tell what are the real dependencies of each elements. I would suggest the same on iOS.

vguerci: Yup that makes sense but I'm afraid this hasn't been followed consistently across the codebase ¯\_(ツ)_/¯

vguerci (approved):

PR #145, opened 2020-11-12
Fix google sdk version

vguerci:
an prepare a patch release


TurpIF (approved):

TurpIF:
This PR is based on `main`. Don't you want to do the hotfix only after the `4.0.1` ? There was new public API meanwhile (contextual)

vguerci:
As no major change has been introduced since the previous version, I don't think this is necessary.

PR #146, opened 2020-11-12
[SKAdNetwork] App Store resolution

vguerci:


CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
```diff
@@ -102,12 +111,16 @@ - (void)URLSession:(NSURLSession *)session
...
     return;
   }
-  CR_URLResolution *resolution = [CR_URLResolution resolutionWithType:CR_URLResolutionStandardUrl];
-  resolution.URL = task.currentRequest.URL;
-  self.resolutionHandler(resolution);
+
+  if (!self.resolved) {
+    CR_URLResolution *resolution = [CR_URLResolution resolutionWithType:CR_URLResolutionStandardUrl
+                                                                    URL:task.currentRequest.URL];
+    self.resolutionHandler(resolution);
```
TurpIF, line 135: Souldn't you call the new `resolveWithResolution` method ?

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
```diff
@@ -24,11 +24,12 @@
 
 @interface CR_URLResolution ()
 @property(nonatomic, strong) NSURL *URL;
-+ (instancetype)resolutionWithType:(CR_URLResolutionType)type;
++ (instancetype)resolutionWithType:(CR_URLResolutionType)type URL:(NSURL *)URL;
 + (instancetype)resolutionError:(NSError *)error;
 @end
 
 @interface CR_URLResolver () <NSURLSessionDataDelegate>
+@property(atomic) BOOL resolved;
```
TurpIF, line 32: Could you explain in the commit message why you want the resolutionHandler to be called only once ?
I guess that there might be a good reason. If so, what do you think about showing it with a test ?

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
```diff
@@ -69,11 +69,24 @@ - (void)resolverURL:(NSURL *)url
...
+  if ([self resolveURL:url]) {
+    return;
+  }
+
   CR_URLRequest *request = [CR_URLRequest requestWithURL:url deviceInfo:deviceInfo];
   NSURLSessionDataTask *task = [self.urlSession dataTaskWithRequest:request];
   [task resume];
 }
 
+- (BOOL)resolveURL:(NSURL *)URL {
```
TurpIF, line 81: I don't understand the naming of this method and I'm confused when reading the code because of the name.

It's trying to resolve AppStore URL if it is an actual AppStore URL.
What do you think about something like `tryResolveAppStoreURL` ?

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
```diff
@@ -66,6 +73,21 @@ - (void)resolverURL:(NSURL *)url
   [task resume];
 }
 
++ (BOOL)isAppStoreURL:(NSURL *)URL {
```
TurpIF, line 90: There is only a `hasSuffix` and `hasPrefix` but honestly this is already a mental gym to just see that you're looking for URL like: `^(apps|itunes|books|music)\..*\.apple\.com$`.

Regex are difficult to read when they are complex. But in this case, I personally think that the regex is simpler than a combination of prefix/suffix checks.

TurpIF (changes requested):

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: Fixed

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: Done

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: Fixed

CriteoPublisherSdk/Sources/Util/CR_URLResolver.m
vguerci: Some people would argue that regexes are evil in all cases, I won't, I think this is mostly a matter of personal taste.
One valid argument would be that some day, adding some rule, could easily defeat the regex being simple.
In any cases, I do not think this is worth the time rewriting this piece of code.

TurpIF (approved):

vguerci:
Thanks!

PR #147, opened 2020-11-12
Merge contexts and put them in CDB request

TurpIF:
JIRA: EE-1324

### DIFF
--- a/CriteoPublisherSdk/Sources/Network/Serializers/CR_BidRequestSerializer.m
+++ b/CriteoPublisherSdk/Sources/Network/Serializers/CR_BidRequestSerializer.m
@@ -26,10 +26,15 @@
 #import "CR_GdprSerializer.h"
 #import "CR_IntegrationRegistry.h"
 #import "CRContextData+Internal.h"
+#import "CR_UserDataHolder.h"
+#import "CRUserData+Internal.h"
+#import "CR_InternalContextProvider.h"
 
 @interface CR_BidRequestSerializer ()
 
 @property(strong, nonatomic, readonly) CR_GdprSerializer *gdprSerializer;
+@property(strong, nonatomic, readonly) CR_UserDataHolder *userDataHolder;
+@property(strong, nonatomic, readonly) CR_InternalContextProvider *internalContextProvider;
 
 @end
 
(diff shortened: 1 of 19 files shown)


CriteoPublisherSdk/Tests/UnitTests/Network/CR_BidRequestSerializerSwiftTests.swift
```diff
@@ -136,12 +138,27 @@ class CR_BidRequestSerializerSwiftTests: XCTestCase {
...
+    XCTAssertEqual(publisher["ext"], expected)
+  }
+
   private func generateBody() -> [String: AnyHashable] {
-    return serializer.body(
-      with: request,
-      consent: consent,
-      config: config,
-      deviceInfo: deviceInfo) as! [String: AnyHashable]
+    return serializer.body(with: request, consent: consent, config: config, deviceInfo: deviceInfo, context: context) as! [String: AnyHashable]
```
vguerci, line 164: Is this the result of `bundle exec fastlane format`? If not, could you please follow the repository formatting?

CriteoPublisherSdk/Sources/Network/Serializers/CR_BidRequestSerializer.m
```diff
@@ -123,4 +123,62 @@ - (NSArray *)slotsWithCdbRequest:(CR_CdbRequest *)cdbRequest {
   return slots;
 }
 
++ (NSDictionary<NSString *, id> *)mergeToNestedStructure:
```
vguerci, line 143: This code depending on client data, which itself could rely on some database or else.
That could easily crash once in production, breaking any bid attempt as I see it will later be used by the `ApiHandler`.
I would strongly suggest to `@try` / `@catch` there with logs / degraded mode fallback value.

vguerci (changes requested):

CriteoPublisherSdk/Tests/UnitTests/Network/CR_BidRequestSerializerSwiftTests.swift
TurpIF: My swift-format has error unfortunately :/ It's weird because I've just update it via brew.
Let me reformat this.

CriteoPublisherSdk/Sources/Network/Serializers/CR_BidRequestSerializer.m
TurpIF: Values are just copied, so even if there is garbage, I don't see why it could fail. But ok, I've done it

CriteoPublisherSdk/Tests/UnitTests/Network/CR_BidRequestSerializerSwiftTests.swift
vguerci: Ah I guess swift-format is still not updated for Xcode 12, probably a reason why I guess I didn't enforce it on CI.
Forget it then, I'll reformat the repo at some point

vguerci (approved):

PR #154, opened 2020-11-30
[Contextual] Implement internal context provider

vguerci:
JIRA: EE-1315

### DIFF
--- a/CriteoPublisherSdk/Sources/Context/CR_InternalContextProvider.m
+++ b/CriteoPublisherSdk/Sources/Context/CR_InternalContextProvider.m
@@ -20,9 +20,31 @@
 #import "CR_InternalContextProvider.h"
 
 #import <sys/utsname.h>
+#import <CoreTelephony/CTTelephonyNetworkInfo.h>
+
+#import "CR_Reachability.h"
+
+@interface CR_InternalContextProvider ()
+@property(nonatomic, strong) CR_Reachability *reachability;
+@end
 
 @implementation CR_InternalContextProvider
 
+#pragma mark - Lifecycle
+
+- (instancetype)init {
+  if (self = [super init]) {
+    self.reachability = [CR_Reachability reachabilityForInternetConnection];
+    [self.reachability startNotifier];
+  }
+  return self;
+}
+
+- (void)dealloc {
+  [self.reachability stopNotifier];
+}
+
+#pragma mark - Public methods
 
 - (nullable NSString *)fetchDeviceMake {
   return @"Apple";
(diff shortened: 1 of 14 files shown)


CriteoPublisherSdk/Sources/Context/CR_InternalContextProvider.m
```diff
@@ -19,14 +19,26 @@
...
 #import "CR_InternalContextProvider.h"
 
+#import <sys/utsname.h>
+
 @implementation CR_InternalContextProvider
 
+
 - (nullable NSString *)fetchDeviceMake {
-  return nil;  // TODO EE-1315
+  return @"Apple";
```
TurpIF, line 53: image

CriteoPublisherSdk/Sources/Context/CR_InternalContextProvider.m
```diff
@@ -177,4 +186,8 @@ - (void)setNonNullObject:(id)object
   return components;
 }
 
+- (CGSize)screenSize {
```
TurpIF, line 192: Isn't the sceenSize and the orientation already available with DeviceUtil ?
AFAIR there is some tooling on DeviceUtil to make it testable.

TurpIF (changes requested):
Do you mind adding few tests on this ?
Value is quite difficult to test, but you can at least check some properties:
- country/language is not null and has length = 2 or 3
- screen size > 0
- orientation is either "Portrait" or "Landscape"
- network is something and does not throw
- model name is not empty

CriteoPublisherSdk/Sources/Context/CR_InternalContextProvider.m
vguerci: I think this needs a test ;p

vguerci:
> Do you mind adding few tests on this ?
> Value is quite difficult to test, but you can at least check some properties:
> 
> * country/language is not null and has length = 2 or 3
> * screen size > 0
> * orientation is either "Portrait" or "Landscape"
> * network is something and does not throw
> * model name is not empty

This is mostly testing Apple's iOS APIs, I'm not convinced of the value of such tests, but let's do it ¯\_(ツ)_/¯

CriteoPublisherSdk/Sources/Context/CR_InternalContextProvider.m
vguerci: AFAIK, There is no `DeviceUtil`?
If you are talking about `DeviceInfo`, I don't see the added value of having such (complex) dependency for a simple case like this?

TurpIF: The added value is to have a single tested/mockable entry point to compute such value.
I may be bias because of Android: there is really a lot of way to compute the screen size: with/without margin bar, rotation, ...

Amokrane:
Are you planning to add some tests in this PR?

CriteoPublisherSdk/Sources/Context/CR_Session.m
```diff
@@ -0,0 +1,43 @@
+//
+//  CR_Session.m
```
Amokrane, line 2: I recommend adding unit tests for this class since it has a bit of logic.

Amokrane (changes requested):

Amokrane:
Even if most of this code calls Apple APIs, it is still worth (in some cases) adding some tests. At least, these tests verify that the right APIs are being called. My rule of thumb when testing code that merely delegates to SDK APIs is to see whether the test is: if it's a low hanging fruit then I write it.
Also there are examples where some methods do have a bit of logic. For example: `getPreferredLanguagesComponentForLocaleKey: https://github.com/criteo/ios-publisher-sdk/pull/154/files#diff-getPreferredLanguagesComponentForLocaleKey b138a9820371d3ed468d90ab28b8d29497a71438e494ebd6ea81e56f00b72529R179

vguerci:
> but let's do it ¯\_(ツ)_/¯
[Done](https://github.com/criteo/ios-publisher-sdk/pull/154/commits/6a955aab0d25457ba648774ebb330e0d00abd03f).

CriteoPublisherSdk/Sources/Context/CR_InternalContextProvider.m
vguerci: In this particular case I doubt we want to compute anything, but want the hardware physical size only, don't we?

Amokrane (approved):

TurpIF (approved):

PR #155, opened 2020-12-01
Fix gdpr param key in GUM endpoint

Amokrane:


CriteoPublisherSdk/Sources/Network/CR_ApiQueryKeys.m
```diff
@@ -85,7 +85,7 @@ + (NSString *)feedbacks {
   return @"feedbacks";
 }
 + (NSString *)gdpr {
-  return @"gdprConsent";
+  return @"gdprString";
```
TurpIF, line 88: Could you rename the name of the method to match the string. So, with the diff we can see all the impacts this change has.

Also, I'm wondering what it the purpose of this file.

TurpIF (changes requested):

CriteoPublisherSdk/Sources/Network/CR_ApiQueryKeys.m
Amokrane: Sure, will do.

Amokrane: Actually, it is the key for the bid API, so there is some side effect (good catch). Let me fix that.

vguerci (approved):

TurpIF (approved):

PR #156, opened 2020-12-03
[Logging] Console debug

vguerci:


CriteoPublisherSdk/Tests/Utility/NetworkManager/CR_NetworkCaptor.m
```diff
@@ -126,27 +127,29 @@ - (void)postToUrl:(NSURL *)url
...
-                         responseHandler(data, error);
-                       }
-                       [self.internalFinishedRequests addObject:content];
-                       [self.internalPendingRequests removeObject:requestContent];
-                       if (self.responseListener != nil) {
-                         self.responseListener(content);
-                       }
-                     }
-                   }];
+    postData = [self.networkManager postToUrl:url
```
Amokrane, line 130: `postedData`

CriteoPublisherSdk/Sources/Network/CR_NetworkManager.h
```diff
@@ -42,10 +42,16 @@ typedef void (^CR_NMResponse)(NSData *_Nullable data, NSError *_Nullable error);
...
-// Assumes all POST calls are made via JSON
-- (void)postToUrl:(NSURL *)url
-           postBody:(NSDictionary *)postBody
-    responseHandler:(nullable CR_NMResponse)responseHandler;
+/**
+ * Send POST request with JSON body
+ * @param url Request URL
+ * @param postBody Dictionary that will be posted as JSON
+ * @param responseHandler Response handler
+ * @return The data posted
```
Amokrane, line 50: Feels a bit weird to return to the caller a value that it (the caller) has already knowledge of.  I know there is some legacy here and I understand that `postToUrl` is handling serialization, but I wanted to get your perspective on how this could be improved in the future.

CriteoPublisherSdk/Sources/Network/CR_ApiHandler.m
```diff
@@ -138,26 +139,33 @@ - (void)doCdbApiCall:(CR_CacheAdUnitArray *)adUnits
...
+                           CRLogInfo(@"BidResponse", @"⬇️ Error: %@", error);
+                           if (completionHandler) {
+                             completionHandler(cdbRequest, nil, error);
+                           }
+                         }
+                         for (CR_CacheAdUnit *adUnit in adUnitChunk) {
+                           [self.bidFetchTracker clearBidFetchInProgressForAdUnit:adUnit];
+                         }
+                       }];
+    if (postData) {
```
Amokrane, line 165: In what case can `postData` be nil?

CriteoPublisherSdk/Sources/Logging/CR_LogMessage.m
```diff
@@ -0,0 +1,96 @@
...
+    case CR_LogSeverityWarning:
+      return @"🟠WARN";
+    case CR_LogSeverityInfo:
+      return @"ℹ️INFO";
+    default:
+      return nil;
+  }
+}
+
+- (NSString *)description {
```
Amokrane, line 81: Do you think it's worth covering this method with tests?

CriteoPublisherSdk/Sources/Logging/CR_LogMessage.m
```diff
@@ -0,0 +1,96 @@
...
+    _function = function;
+    _timestamp = [NSDate date];
+    _exception = exception;
+    _message = message;
+  }
+
+  return self;
+}
+
++ (instancetype)messageWithTag:(NSString *)tag
```
Amokrane, line 46: Do you think it's worth covering this method with tests?

Amokrane (changes requested):

CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
 # Criteo Publisher SDK Changelog
+--------------------------------------------------------------------------------
+## Version 4.x.x
+
+### Fixes
+- **In House**: `[Criteo loadBidForAdUnit:responseHandler:]` now call `responseHandler` with nil bid
```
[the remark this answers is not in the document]
vguerci: Done

CriteoPublisherSdk/Sources/Logging/CR_LogMessage.h
```diff
@@ -0,0 +1,63 @@
...
+#define CRLogMessage(TAG, SEVERITY, EXCEPTION, ARGS...) \
+  [CR_LogMessage messageWithTag:TAG                     \
+                       severity:SEVERITY                \
+                           file:__FILE__                \
+                           line:__LINE__                \
+                       function:__PRETTY_FUNCTION__     \
+                      exception:EXCEPTION               \
+                         format:ARGS]
+
+// Note: Values are following syslog protocol: https://www.rfc-editor.org/info/rfc5424
```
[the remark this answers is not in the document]
vguerci: This note explains why these 3-4-6-7 values were picked, and were to get additional one if needed, which is, imo the thing I would like to be explained to me when looking at this code.
I would assume any developper to know what are info / debug and so forth, don't you think so?

CriteoPublisherSdk/Sources/Logging/CR_LogMessage.m
```diff
@@ -0,0 +1,96 @@
...
+                              file:[NSString stringWithUTF8String:file]
+                              line:(NSUInteger)line
+                          function:[NSString stringWithUTF8String:function]
+                         exception:exception
+                           message:message];
+}
+
+- (NSString *)severityLabel {
+  switch (self.severity) {
+    case CR_LogSeverityDebug:
```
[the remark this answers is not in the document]
vguerci: Done

CriteoPublisherSdk/Tests/Utility/NetworkManager/CR_NetworkCaptor.m
vguerci: Done

CriteoPublisherSdk/Sources/Network/CR_NetworkManager.h
vguerci: Clearly this is was the shortest past solution w/o being too intrusive / refactoring many things.
If we had time, I would rewrite most of the network stack of the SDK as the current Caller -> ApiHandler / NetworkManager could deserve some improvements (there are some tickets in the project backlog about this)
A decent middle ground solution could be to add extra arguments to this method to specify that we want to log JSON posted data, as the problem is that we do not want to log all queries, but we want JSON format (TS spec) which is only available there.

CriteoPublisherSdk/Sources/Network/CR_ApiHandler.m
vguerci: On nil data / json serialisation error, which both are error cases that should be handled / logged differently.

CriteoPublisherSdk/Sources/Logging/CR_LogMessage.m
vguerci: I don't think so as it is only implemented for development purposes, and isn't in production code path.
I'd rather remove it than testing.

For information, `-[NSObject description]` is the Objective-C equivalent of Java's `toString()` one, which is informational for for debugging purposes, not worth to test imo most of the time, exceptions being like for the other tests, a client (TS in this case) that request specifically some data to be logged.

CriteoPublisherSdk/Sources/Logging/CR_LogMessage.m
vguerci: I don't think so, variadic parameters can only be consumed this way, which is and old are really low level C APIs...

By the way, I could not figure a way to defer that string interpolation at logging time to spare some CPU cycles when the message isn't ultimately logged.
Maybe when working on remote logging I might remove that class altogether to address this, as you cannot store such variadic parameters / interpolate strings natively with arguments as an array.

Maybe you have some ideas on the topic?

CriteoPublisherSdk/Sources/Logging/CR_LogMessage.h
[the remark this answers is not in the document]
Amokrane: `I would assume any developper to know what are info / debug and so forth, don't you think so?
`
info/debug are probably clear, but the distinction between error & warnings is not that trivial IMHO.

CriteoPublisherSdk/Sources/Network/CR_NetworkManager.h
Amokrane: Yes, it does. 

> A decent middle ground solution could be to add extra arguments to this method to specify that we want to log JSON posted data, as the problem is that we do not want to log all queries, but we want JSON format (TS spec) which is only available there.

It does prevent the need of returning the serialized JSON, however it doesn't allow customizing the log message relative to the type of request being made.

I wonder whether `postBody` can be used for logging?

CriteoPublisherSdk/Sources/Logging/CR_LogMessage.m
Amokrane: Ok, my bad, thought it was returning the value that was used by the logger when printing on the console.

CriteoPublisherSdk/Sources/Network/CR_NetworkManager.h
vguerci: The idea was to have one postBody argument to be used in the log message to address this.

Anyway, for now, don't you think the current simple approach is clean enough until a day where some refactoring can be done in this part of the codebase?

Amokrane: Sure, lgtm.

Amokrane (dismissed):

vguerci:
@Amokrane Thanks for your review, I've since addressed the bug reported by @irwansetiawan and your suggestion about the postData by moving logging logic directly into the network manager, you can see changes [there](https://github.com/criteo/ios-publisher-sdk/compare/8d1064ececdb9b8bc075c49466033432eb9978c8..b2b15b70d55b56ac79408ba905b1b58727fa0817). Could you please review this so I can proceed to release process?

Amokrane (approved):
LGTM

PR #174, opened 2021-01-05
[Logging] Add extensible handler

vguerci:


CriteoPublisherSdk/Sources/Logging/CR_LogHandler.m
```diff
@@ -0,0 +1,116 @@
...
+        _severityThreshold = handler.severityThreshold;
+      }
+    }
+  }
+  return self;
+}
+
+#pragma mark - LogHandler
+
+- (void)setSeverityThreshold:(CR_LogSeverity)threshold {
```
TurpIF, line 94: I'm not sure about the usefulness of the severityThreshold for the multiplexer.

In our case, both handlers are really different: 
* console handler log level is controlled by publisher
* remote handler log level is controlled by remote config

Moreover, the _severityThreshold field is only written in the implementation but never read.

What do you think about making this threshold an implementation detail for both handlers ?

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.m
```diff
@@ -0,0 +1,116 @@
...
+
+- (void)logMessage:(CR_LogMessage *)message {
+  for (id<CR_LogHandler> handler in self.logHandlers) {
+    [handler logMessage:message];
+  }
+}
+
+- (CR_ConsoleLogHandler *)consoleLogHandler {
+  for (id<CR_LogHandler> logHandler in self.logHandlers) {
+    if ([logHandler isKindOfClass:CR_MultiplexLogHandler.class]) {
```
TurpIF, line 109: Do you mean `CR_ConsoleLogHandler.class` instead ?

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.m
```diff
@@ -0,0 +1,116 @@
...
+- (void)setSeverityThreshold:(CR_LogSeverity)threshold {
+  for (id<CR_LogHandler> handler in self.logHandlers) {
+    handler.severityThreshold = threshold;
+  }
+  _severityThreshold = threshold;
+}
+
+- (void)logMessage:(CR_LogMessage *)message {
+  for (id<CR_LogHandler> handler in self.logHandlers) {
+    [handler logMessage:message];
```
TurpIF, line 103: To make each handlers independent, what do you think about try/catch this ?

TurpIF (changes requested):

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.h
```diff
@@ -0,0 +1,42 @@
...
+@interface CR_ConsoleLogHandler : NSObject <CR_LogHandler>
+@end
+
+@interface CR_ConsoleLogHandler (Testing)
+- (void)logMessageToConsole:(CR_LogMessage *)logMessage;
+@end
+
+@interface CR_MultiplexLogHandler : NSObject <CR_LogHandler>
+@property(nonatomic, readonly) NSArray<id<CR_LogHandler>> *logHandlers;
+- (instancetype)initWithLogHandlers:(NSArray *)logHandlers;
```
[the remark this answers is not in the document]
vguerci: Should work, but I would suggest to do this after implementing the remote log handler.

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.m
vguerci: Done.

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.m
vguerci: Has been fixed by a later change.

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.m
vguerci: Done.

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.h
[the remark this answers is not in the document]
TurpIF: Ok to keep in this way for now.
Meanwhile, a test could show (and guard) if a workaround is necessary:
* Given SDK is not initialized
* When logging something
* Then no exception is thrown
* And a log is printed in console
* And a log is pushed in remote log storage (optional)

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.m
```diff
@@ -0,0 +1,103 @@
...
+
+@implementation CR_ConsoleLogHandler
+
+@synthesize severityThreshold = _severityThreshold;
+
+#pragma mark - Lifecycle
+
+- (instancetype)init {
+  if (self = [super init]) {
+    self.severityThreshold = crConsoleSeverityThresholdDefault;
```
TurpIF, line 32: Question:
Here, your setting the property rather than its backing field `_severityThreshold`.
Do you have a documentation explaining when to set the property, when to set the backing field, and what is the difference ?

TurpIF (dismissed):

CriteoPublisherSdk/Sources/Logging/CR_LogHandler.m
vguerci: The coding style we are following has a [chapter about this topic](https://github.com/raywenderlich/objective-c-style-guide#underscores):
> An exception to this: inside initializers, the backing instance variable (i.e. _variableName) should be used directly to avoid any potential side effects of the getters/setters.

Amending then.

TurpIF (dismissed):

PR #178, opened 2021-01-07
[Actions] Fix pr test

vguerci:


TurpIF (changes requested):
Tests are run twice in the PR. Maybe you should just remove the `branches: ['*']` under the `pull_requests` section ?

vguerci:
> Maybe you should just remove the `branches: ['*']` under the `pull_requests` section ?

This is already the case, isn't? Basically removing that is the fix.
The other addition is to be aligned on what we do on main branch, i.e run tests once merged.

This double run only occurs on this PR, I guess the test on push is to be ignored and a consequence of me pushing on the base branch accidentally earlier.

TurpIF (approved):

PR #182, opened 2021-01-12
Disable RemoteLogs when consent is not given

TurpIF:
JIRA: EE-1374


vguerci (dismissed):
Doesn't this only partially implement the task. Shouldn't that also clean queue / prevent send?

TurpIF:
> Doesn't this only partially implement the task. Shouldn't that also clean queue / prevent send?

This prevent logs to be gathered at log time. This means that if there is something in the queue, then the consent was given. So I would say that we have the right to send it.

Moreover, I guess that the main transition we will face is: no consent -> got consent. In such transition, there is no issue.

vguerci (approved):
Ok, NIT: maybe just edit the ticket description accordingly for the record

PR #184, opened 2021-01-12
Add serializer for remote log records

TurpIF:
JIRA: EE-1350

### DIFF
--- a/CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.h
+++ b/CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.h
@@ -1,0 +1,32 @@
+//
+//  CR_RemoteLogRecordSerializer.h
+//  CriteoPublisherSdk
+//
+//  Copyright © 2018-2021 Criteo. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+#import "CR_RemoteLogRecord.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CR_RemoteLogRecordSerializer : NSObject
+
+- (NSArray<NSDictionary<NSString *, NSObject *> *> *)serializeRecords:
+    (NSArray<CR_RemoteLogRecord *> *)records;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
+++ b/CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
@@ -1,0 +1,68 @@
+//
+//  CR_RemoteLogRecordSerializer.m
+//  CriteoPublisherSdk
+//
+//  Copyright © 2018-2021 Criteo. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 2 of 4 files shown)


CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
```diff
@@ -0,0 +1,67 @@
...
+    NSDictionary<NSString *, NSObject *> *dictionary = [self dictionaryForRemoteLogRecord:record];
+    [formattedArray addObject:dictionary];
+  }
+  return formattedArray;
+}
+
+- (NSDictionary<NSString *, NSObject *> *)dictionaryForRemoteLogRecord:
+    (CR_RemoteLogRecord *)record {
+#pragma clang diagnostic push
+#pragma ide diagnostic ignored "OCLegacyObjCLiteralInspection"
```
vguerci, line 37: NIT: If this very specific to AppCode I would suggest removing it as the comment is quite explicit. Maybe also add a warning about adding trailing entries using this constructor or maybe use `dictionaryWithObjects:forKeys:` instead.

CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
```diff
@@ -0,0 +1,67 @@
...
+
+- (NSString *)formatSeverity:(CR_LogSeverity)severity {
+  switch (severity) {
+    case CR_LogSeverityError:
+      return @"Error";
+    case CR_LogSeverityWarning:
+      return @"Warning";
+    case CR_LogSeverityInfo:
+      return @"Info";
+    default:
```
vguerci, line 62: NIT: I would rather avoid having a default there to enforce a developer to amend this if levels were to be added. Unless there is a consequence on backend.

vguerci (approved):

CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
TurpIF: If level is not known by backend (Debug, Info, Warning or Error), then the backend will just throw out the whole request. Not the best behavior, but at least, it will not make the server throw.

Let me remove the default then.

CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
TurpIF: Ok let me remove the `OCLegacyObjCLiteralInspection`.

And you're right, `dictionaryWithObjects:forKeys:` seems better to use when handling `nil` values.

CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.h
```diff
@@ -0,0 +1,32 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import "CR_RemoteLogRecord.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface CR_RemoteLogRecordSerializer : NSObject
+
+- (NSArray<NSDictionary<NSString *, NSObject *> *> *)bodyForRemoteLogRecords:
```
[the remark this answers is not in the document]
TurpIF: Ok for `serializeRecords`

CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
TurpIF: In fact `dictionaryWithObjects:forKeys:` needs an array. For the same reason, `@[]` notation can't be used. So I need to use `+ (instancetype)arrayWithObjects:(ObjectType)firstObj, ... NS_REQUIRES_NIL_TERMINATION;` which has the same issue with `nil`.
So instead, I will use a mutable dictionary if it's ok for you.

vguerci: As you wish, this is just NIT, the former version is fine as well :)

CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
```diff
@@ -0,0 +1,68 @@
...
+  NSMutableArray *formattedArray = [NSMutableArray arrayWithCapacity:records.count];
+  for (CR_RemoteLogRecord *record in records) {
+    NSDictionary<NSString *, NSObject *> *dictionary = [self serializeRecord:record];
+    [formattedArray addObject:dictionary];
+  }
+  return formattedArray;
+}
+
+- (NSDictionary<NSString *, NSObject *> *)serializeRecord:(CR_RemoteLogRecord *)record {
+  // The exception type is nullable, so it is not possible to use @{} notation.
```
vguerci, line 35: NIT: this is something I'm not asking you to do, but you could init that mutable dict with a literal ;)

vguerci (approved):

CriteoPublisherSdk/Sources/Logging/CR_RemoteLogRecordSerializer.m
TurpIF: could you show me the syntax ?

PR #185, opened 2021-01-13
[Logging] Send logs to backend

vguerci:


TurpIF (dismissed):
LGTM, although, there is no tests, but I'm wondering if it is really worth it in such case. I let you decide on this.

vguerci:
> LGTM, although, there is no tests, but I'm wondering if it is really worth it in such case. I let you decide on this.

Basically I've based this on the `sendCsm` method which is untested and was wondering if that was worth testing, anyway, [added a couple of tests](https://github.com/criteo/ios-publisher-sdk/compare/33de381c666ae153d737b29a5460f0c5210bba54..91482d7a289968ef46282a36750f82e4b2a02328).

TurpIF (approved):

vguerci:
Thanks!

PR #204, opened 2021-02-09
[Network][Config] Send device OS and Model

vguerci:
So we can configure per os

JIRA: DPP-3177


TurpIF (approved):
I'm curious why you need the device model too ? Do you have a use case in mind ?

vguerci:
> I'm curious why you need the device model too ? Do you have a use case in mind ?

Nope, just replicated what is done on another endpoint, thinking that it could serve for some purpose one day.
It isn't read by backend change so far.

TurpIF (approved):

PR #206, opened 2021-02-24
[Logging] Improve logging

vguerci:
- Log network errors as info (instead of warnings)
- Do not log localizedDescription
- Do not parse empty responses

JIRA: DPT-3233


TurpIF (approved):

vguerci:
Thank you

PR #210, opened 2021-03-15
Google Ads SDK v8 support: Fix adapter podspec dependency version

vguerci:


CriteoPublisherSdk.podspec
```diff
@@ -39,7 +39,7 @@ Pod::Spec.new do |spec|
   spec.subspec "GoogleAdapter" do |adapter|
     adapter.source_files     = "CriteoGoogleAdapter/Sources/**/*.{h,m}"
     adapter.dependency         "CriteoPublisherSdk/Sdk"
-    adapter.dependency         "Google-Mobile-Ads-SDK", "~> 7.49"
+    adapter.dependency         "Google-Mobile-Ads-SDK", ">= 7.49", "<9"
```
TurpIF, line 42: Why the <9 ? Do you think we will definitely be impacted by a breaking change ?

TurpIF (dismissed):

CriteoPublisherSdk.podspec
vguerci: That is explicitly what we support :)

TurpIF (approved):

PR #234, opened 2021-11-23
Rewarded AdUnits

CedricGatay:
Introduce rewarded ad units to SDK:

 * it is only available to GAM integration, using it with another one will simply ignore it and log the thing
 * cdb calls now use the `isRewarded` flag and parse the `rewarded` response in slots.

Side note: bumped dependencies before starting development of this branch

Jira: DPP-3644 / DPP-3645


CriteoPublisherSdk/Sources/CR_AdUnitHelper.m
```diff
@@ -52,6 +52,10 @@ + (CR_CacheAdUnit *)cacheAdUnitForAdUnit:(CRAdUnit *)adUnit {
       return [[CR_CacheAdUnit alloc] initWithAdUnitId:adUnit.adUnitId
                                                  size:nativeSize
                                            adUnitType:CRAdUnitTypeNative];
+    case CRAdUnitTypeRewarded:
+      return [[CR_CacheAdUnit alloc] initWithAdUnitId:adUnit.adUnitId
+                                                 size:nativeSize
```
TurpIF, line 57: For rewarded ad unit, you should use the screen size like fr interstitial above

CriteoPublisherSdk/Sources/Configuration/CR_IntegrationRegistry.m
```diff
@@ -80,4 +81,21 @@ - (BOOL)isAdMobMediationPresent {
          NSProtocolFromString(@"GADCustomEventBanner") != nil;
 }
 
+- (CR_CacheAdUnitArray *)filterAdUnits:(CR_CacheAdUnitArray *)units {
```
TurpIF, line 84: The naming is kind of weird in the context of the CR_IntegrationRegistry.
From this class, I would either expect a method indicating which integrations are supported for the given ad unit (or a predicate taking an ad unit corresponding of your block), or a method that keep only supported ad units.

Yours is doing that but the naming is not perfect IMHO. What do you think about `filterSupportedAdUnits` or `keepSupportedAdUnits`, or `isAdUnitSupported` (for predicate)

CriteoPublisherSdk/Sources/Configuration/CR_IntegrationRegistry.m
```diff
@@ -80,4 +81,21 @@ - (BOOL)isAdMobMediationPresent {
...
+  return [units
+      filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(id _Nullable adUnit,
+                                                                        NSDictionary<NSString *, id>
+                                                                            *_Nullable bindings) {
+        if ([adUnit adUnitType] != CRAdUnitTypeRewarded) {
```
TurpIF, line 89: In my opinion, listing exclusion is more easy that listing inclusion. For instance:

```
if (adunit is rewarded and profileId != GAM) {
  log("warning")
  return false;
}

// maybe a future exclusion ?

return true;
```

Moreover, it will then be much easier to add other exclusions then.

CriteoPublisherSdk/Sources/Configuration/CR_IntegrationRegistry.m
```diff
@@ -80,4 +81,21 @@ - (BOOL)isAdMobMediationPresent {
...
+          return YES;
+        }
+        if ([self.profileId isEqual:@(CR_IntegrationGamAppBidding)]) {
+          return YES;
+        } else {
+          CRLogWarn(@"Bidding", @"RewardedAdUnit %@ can only be used with GAM app bidding",
```
TurpIF, line 95: Do you mind adding a extra comment explaining the why of such filter ?

The why:
> OnlyGAM AppBidding is supporting rewarded ads because they are handling the display themselves.

CriteoPublisherSdk/Sources/Network/CR_ApiHandler.m
```diff
@@ -121,7 +121,8 @@ - (void)doCdbApiCall:(CR_CacheAdUnitArray *)adUnits
               context:(CRContextData *)contextData
         beforeCdbCall:(CR_BeforeCdbCall)beforeCdbCall
     completionHandler:(CR_CdbCompletionHandler)completionHandler {
-  CR_CacheAdUnitArray *requestAdUnits = [self filterRequestAdUnitsAndSetProgressFlags:adUnits];
+  CR_CacheAdUnitArray *filteredAdUnits = [self.integrationRegistry filterAdUnits:adUnits];
```
TurpIF, line 124: This filter is done in a quite low level here. Isn't it possible to move it a little bit before.

I see one (weird) case which might provoke a bug:
- given CDB answering a bid
- given the SDK storing the bid in the cache as a rewarded ad unit (either because CDB wrongly answered or because of an adjacent bug in the SDK)
- when getting a next bid
- then bid in cache is retrieved
- then a rewarded bid is used while it's not usable

Of course we don't want the backend to answer crap, but we never know. For instance for native requests CDB was not always answering native bids before a bug was fixed.

CriteoPublisherSdk/Tests/UnitTests/Configuration/CR_IntegrationRegistryTests.m
```diff
@@ -143,4 +143,35 @@ - (void)testIsAdMobMediationPresent_GivenAdMobMediationNotInRuntime_ReturnNo {
   XCTAssertFalse(isPresent);
 }
 
-@end
\ No newline at end of file
+- (void)testFilterAdUnits_GivenProfileIdNotGAM_RemoveRewardedAdUnit {
```
TurpIF, line 146: Do you mind adding a test case in InHouse functional tests ? It a little bit a duplication of your test here, but it's to keep a layer of functional tests that shows the features and boundary of the SDK

CriteoPublisherSdk/Sources/CRRewardedAdUnit.m
```diff
@@ -29,7 +29,7 @@ - (instancetype)initWithAdUnitId:(NSString *)adUnitId {
 }
 
 - (NSUInteger)hash {
-  return super.hash ^ (NSUInteger)14559042078869117629ull;
+  return super.hash ^ (NSUInteger)14559065388869353629ull;
```
TurpIF, line 32: Just for my understanding: why and how did you get this value (prime number ?) ? I see that `self.class` is part of the equality, but the hash can be surjective, so why bothering with such number. Moreover, the parent class already contains the ad unit type which kind of of represents the class.

CedricGatay: I just applied the same logic than the one was in place with other Ad unit types, without thinking that much I admit.

TurpIF (approved):
LGTM, don't hesitate to ping me back once test are green t re-review you PR.

CedricGatay:
Tests are bugging, tried various things but they are still annoying, digging further...

CriteoPublisherSdk/Tests/IntegrationTests/CR_BidManagerIntegrationTests.m
```diff
@@ -80,15 +78,13 @@ - (void)tearDown {
...
-                 withContext:self.contextData];
+- (void)test_given2AdUnits_whenPrefetchBid_thenGet2Bids {
+  [self.bidManager prefetchBidsForAdUnits:@[ self.cacheAdUnit1, self.cacheAdUnit2 ]
+                              withContext:self.contextData];
 
-  [self _waitNetworkCallForBids:@[ self.cacheAdUnit1, self.cacheAdUnit2, self.cacheAdUnit3 ]];
+  [self _waitNetworkCallForBids:@[ self.cacheAdUnit1, self.cacheAdUnit2 ]];
   XCTAssertNotNil([self.bidManager getBidThenFetch:self.cacheAdUnit1 withContext:self.contextData]);
   XCTAssertNotNil([self.bidManager getBidThenFetch:self.cacheAdUnit2 withContext:self.contextData]);
-  XCTAssertNotNil([self.bidManager getBidThenFetch:self.cacheAdUnit3 withContext:self.contextData]);
```
TurpIF: Why do you need to remove such test ?

CedricGatay: Would have needed to mock static injection of `integrationRegistry` inside `AdUnitHelper` which should be easier to do once refactored on the other end. 

I did not want to make this PR last for too long, but if you consider the test relevant, I can work this out.

TurpIF: The `AdUnitHelper` dependency is fetch statically, but still, it is fetch during the `bidManager getBidThenFetch` call. So you can mock the AdUnitHelper in the static dependency provider meantime.

I see that this test (https://github.com/criteo/ios-publisher-sdk/pull/234/files#diff-df5582f92dc8c24b9c4a3a4e05d75ddad728a4fa10babedeba1e89c3d7ae0f1aR200) is verifying that the feature don't work on InHouse as expected.

But I don't see any functional test checking that it is working for DFP. Do you mind adding such tests (like `DfpBannerFunctionalTests.m`) testing that, without mocks (except the network), it is possible to get a bid response for DFP integration (no need to go E2E with DFP ad server, but you're welcome if you want to).

CedricGatay: I've updated the content to allow these tests.

As we're relying on global shared object I need to set the value on this same static object, when `AdUnitHelper` refactor will be completed we can refactor this to benefit from IoC.

Also added a `CR_DfpRewardedFunctionalTests.m` file that does the opposite of the `InHouseFunctionalTest` checking that we're issuing http requests for Rewarded Ads.

CriteoPublisherSdk/Tests/IntegrationTests/CR_BidManagerIntegrationTests.m
```diff
@@ -67,6 +69,9 @@ - (void)setUp {
   self.adUnit1 = [CR_TestAdUnits preprodBanner320x50];
   self.cacheAdUnit1 = [CR_AdUnitHelper cacheAdUnitForAdUnit:[CR_TestAdUnits preprodBanner320x50]];
   self.cacheAdUnit2 = [CR_AdUnitHelper cacheAdUnitForAdUnit:[CR_TestAdUnits preprodInterstitial]];
+  [Criteo.sharedCriteo.dependencyProvider.integrationRegistry declare:CR_IntegrationGamAppBidding];
```
TurpIF, line 72: It might be worth adding a specific TODO with a new jira ticket (like TODO DPP-11234) so you start tracing all the places that will eventually need to be corrected when AdUnitHelper get refactored correctly.

TurpIF (changes requested):

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
```diff
@@ -0,0 +1,69 @@
...
+@end
+
+@implementation CR_DfpRewardedFunctionalTests
+
+
+#pragma mark - rewarded
+
+-(void)setUp{
+    // WARNING AdUnitHelper does use singleton object (no DI in here)
+  [Criteo.sharedCriteo.dependencyProvider.integrationRegistry declare:CR_IntegrationGamAppBidding ];
```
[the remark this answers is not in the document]
CedricGatay: I don't think I get this one, I need to force the static provider to be able to test what I want, am I right ?

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
```diff
@@ -49,18 +51,34 @@ - (void)tearDown {
...
+  XCTAssertEqualObjects(targeting[CR_TargetingKey_crtCpm], @"1.12");
+  XCTAssertEqual(targeting[CR_TargetingKey_crtFormat], @"video");
+  XCTAssertEqualObjects(targeting[CR_TargetingKey_crtSize], @"320x480");
+
+  // as this is a "video" url is not base64ed
+  NSString *encodedUrl = targeting[CR_TargetingKey_crtDfpDisplayUrl];
+  NSString *decodedUrl =
+      [[encodedUrl stringByRemovingPercentEncoding] stringByRemovingPercentEncoding];
+
+  XCTAssertEqualObjects(bid.displayUrl, decodedUrl);
```
TurpIF, line 79: It's weird to assert an actual coming from production code with an expected also coming from production code.
But I saw that it's done the same in other functional tests so ¯\_(ツ)_/¯

Ok for now as you're also testing that there is a cpm, a format, a size and that, at least, the display URL is not null.

TurpIF (approved):

PR #236, opened 2021-12-13
Rewarded ad in showcase app

CedricGatay:
Added rewarded ads in showcase app

Reworked the way integration are detected  when enriching requests. It no longer fails when the bid can not be retrieved properly.

### DIFF
--- a/CriteoAdViewer/Sources/AdViewer/AdConfig.swift
+++ b/CriteoAdViewer/Sources/AdViewer/AdConfig.swift
@@ -20,11 +20,13 @@
 struct AdConfig {
   let publisherId: String
   let adUnit: CRAdUnit
+  let externalAdUnitId: String
   let adFormat: AdFormat
 
-  init(publisherId: String, adUnitId: String, adFormat: AdFormat) {
+  init(publisherId: String, adUnitId: String, externalAdUnitId: String? = nil, adFormat: AdFormat) {
     self.publisherId = publisherId
     self.adUnit = AdConfig.buildAdUnit(adFormat: adFormat, adUnitId: adUnitId)
+    self.externalAdUnitId = externalAdUnitId ?? adUnitId
     self.adFormat = adFormat
   }
 
--- a/CriteoAdViewer/Sources/AdViewer/AdConfig.swift
+++ b/CriteoAdViewer/Sources/AdViewer/AdConfig.swift
@@ -36,6 +36,8 @@ struct AdConfig {
       return CRNativeAdUnit(adUnitId: adUnitId)
     case .flexible(.interstitial), .flexible(.video):
       return CRInterstitialAdUnit(adUnitId: adUnitId)
+    case .flexible(.rewarded):
+      return CRRewardedAdUnit(adUnitId: adUnitId)
     case _:
       fatalError("Unsupported")
     }
--- a/CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
+++ b/CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
@@ -35,6 +35,9 @@
 
 @interface CR_DfpRewardedFunctionalTests : CR_IntegrationsTestBase
 
+// to be able to restore
+@property(strong, nonatomic) CR_IntegrationRegistry *oldIntegrationRegistry;
+
 @end
 
 @implementation CR_DfpRewardedFunctionalTests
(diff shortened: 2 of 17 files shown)


CriteoAdViewer/Sources/AdViewer/AdConfig.swift
```diff
@@ -20,11 +20,13 @@
 struct AdConfig {
   let publisherId: String
   let adUnit: CRAdUnit
+  let externalAdUnitId: String
```
TurpIF, line 23: Such property is only used for rewarded. Do you mind updating the methods generating GAM ad request object to use this property instead of the old one ? This should not change any behavior, but this might be wiser to do it now and don't forget about this particularity rather than having to debug this later asking why this doesn't work on native, banner, ...

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
```diff
@@ -43,26 +46,82 @@ @implementation CR_DfpRewardedFunctionalTests
...
+  BOOL finished = [self.criteo testing_waitForRegisterHTTPResponses];
+  XCTAssert(finished, "Failed to receive all prefetch requests");
+}
+
+- (void)test_givenRewarded_GAMIntegrationDeclared_whenLoadTwoTimes_adLoaded {
+  CRRewardedAdUnit *adUnit = [CR_TestAdUnits rewarded];
+
+  self.criteo = [Criteo testing_criteoWithNetworkCaptor];
+  // use same integrationRegistry in test criteo object and global instance (see DPP-3734)
+  CR_IntegrationRegistry *integrationRegistry = [[CR_IntegrationRegistry alloc] init];
```
TurpIF, line 106: Do you mind extracting this into a méthod. So this Jacky duplication get wrapped

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
```diff
@@ -43,26 +46,82 @@ @implementation CR_DfpRewardedFunctionalTests
...
+  // register here fails as integration is not yet detected
   [self.criteo testing_registerWithAdUnits:@[ adUnit ]];
 
   GAMRequest *request = [[GAMRequest alloc] init];
 
+  // this one make the detection logic discover GAM integration
+  [self enrichAdObject:request forAdUnit:adUnit];
+
+  // register is now ok due to GAM being detected
+  [self.criteo testing_registerWithAdUnits:@[ adUnit ]];
```
TurpIF, line 74: This operation should do nothing as the criteo instance was already registered. Or maybe am I missing something ?

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
```diff
@@ -43,26 +46,82 @@ @implementation CR_DfpRewardedFunctionalTests
...
+  CR_IntegrationRegistry *integrationRegistry = [[CR_IntegrationRegistry alloc] init];
+  [integrationRegistry declare:CR_IntegrationGamAppBidding];
+  self.criteo.dependencyProvider.integrationRegistry = integrationRegistry;
+  Criteo.sharedCriteo.dependencyProvider.integrationRegistry = integrationRegistry;
+
+  // GAM is declared so adUnit can be properly built
+  [self.criteo testing_registerWithAdUnits:@[ adUnit ]];
+
+  GAMRequest *request = [[GAMRequest alloc] init];
+
```
TurpIF, line 115: Same here, this should be a noop

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
```diff
@@ -43,26 +46,82 @@ @implementation CR_DfpRewardedFunctionalTests
...
+  // this one make the detection logic discover GAM integration
+  [self enrichAdObject:request forAdUnit:adUnit];
+
+  // register is now ok due to GAM being detected
+  [self.criteo testing_registerWithAdUnits:@[ adUnit ]];
+
+  // this one warms up the cache
+  [self enrichAdObject:request forAdUnit:adUnit];
+
+  // this one passes properly
```
TurpIF, line 79: What do you think about extracting all this logic below in a dedicated méthod ? So this is a test method doing action+expectations, and you have 2 different entry points doing the warmup accordingly.

wiremock/mappings/inapp-v2.json
```diff
@@ -0,0 +1,26 @@
+{
```
TurpIF, line 1: This file is only for vast request. Wasn't there mocked response for other ad units before ? Maybe should you rename this accordingly and makes the preconditions more precise ?

TurpIF (changes requested):

wiremock/mappings/inapp-v2.json
CedricGatay: Updated with a precondition on query to only respond to rewarded ones.

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
CedricGatay: Method extracted to warm and enrich

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
CedricGatay: Removed as it is useless

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
CedricGatay: Useless operation, removed it.

CriteoPublisherSdk/Tests/IntegrationTests/CR_DfpRewardedFunctionalTests.m
CedricGatay: Extracted to mark the code needing adjustments.

CriteoAdViewer/Sources/AdViewer/AdConfig.swift
CedricGatay: Updated usages in `GoogleAdViewBuilder` to make use of this variable instead of ad unit id for future proofing.

TurpIF (approved):
LGTM

PR #237, opened 2021-12-27
feat(DPP-3746): allow empty constructor usage for `CRBannerView`

CedricGatay:
Allow to use empty constructor to init `CRBannerView`

Adjusted behavior so that CRBannerView does not require adUnitId set on init but get it from the bid. Made the sample app use this for inHouse showcase.

Fixes DPP-3746

### DIFF
--- a/CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
+++ b/CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
@@ -38,6 +38,10 @@
 
 @implementation CRBannerView
 
+- (instancetype)init {
+  return [self initWithAdUnit:[CRBannerAdUnit alloc]];
+}
+
 - (instancetype)initWithAdUnit:(CRBannerAdUnit *)adUnit {
   return [self initWithAdUnit:adUnit criteo:[Criteo sharedCriteo]];
 }
@@ -159,11 +163,16 @@
     return;
   }
 
-  CR_CdbBid *cdbBid = bid.consume;
+  CR_CdbBid *cdbBid = [bid consumeFor:CRAdUnitTypeBanner];
   if (!cdbBid) {
     [self safelyNotifyAdLoadFail:CRErrorCodeNoFill];
     return;
   }
+
+  CGFloat width = [cdbBid.width floatValue];
+  CGFloat height = [cdbBid.height floatValue];
+  self.frame = CGRectMake(0, 0, width, height);
+  self.webView.frame = CGRectMake(0, 0, width, height);
 
   [self loadAdWithCdbBid:cdbBid];
 }
--- a/CriteoPublisherSdk/Sources/Standalone/CRInterstitial.m
+++ b/CriteoPublisherSdk/Sources/Standalone/CRInterstitial.m
@@ -59,6 +59,10 @@
   return self;
 }
 
+- (instancetype)init {
+  return [self initWithAdUnit:[CRInterstitialAdUnit alloc]];
+}
+
 - (instancetype)initWithAdUnit:(CRInterstitialAdUnit *)adUnit {
   return [self initWithAdUnit:adUnit criteo:[Criteo sharedCriteo]];
 }
@@ -177,7 +181,7 @@
     return;
   }
 
-  CR_CdbBid *cdbBid = bid.consume;
+  CR_CdbBid *cdbBid = [bid consumeFor:CRAdUnitTypeInterstitial];
   if (!cdbBid) {
     [self safelyNotifyAdLoadFail:CRErrorCodeNoFill];
     self.isAdLoading = NO;
--- a/CriteoPublisherSdk/Tests/Utility/Checkers/CR_CreativeViewChecker.h
+++ b/CriteoPublisherSdk/Tests/Utility/Checkers/CR_CreativeViewChecker.h
@@ -28,6 +28,8 @@
 
 - (instancetype)init NS_UNAVAILABLE;
 
+- (instancetype)initWithCriteo:(Criteo *)criteo;
+
 - (instancetype)initWithAdUnit:(CRAdUnit *)adUnit criteo:(Criteo *)criteo;
 
 @property(strong, nonatomic, readonly) UIWindow *uiWindow;
@@ -44,9 +46,9 @@
 
 - (void)resetBannerView;
 
-- (void)injectBidWithExpectedCreativeUrl:(NSString *)creativeUrl;
+- (void)injectBidWithExpectedCreativeUrl:(NSString *)creativeUrl forAdUnit:(CRAdUnit *)adUnit;
 
-- (void)injectBidWithAppStoreClickUrl;
+- (void)injectBidWithAppStoreClickUrl:(CRAdUnit *)adUnit;
 - (void)clickUrl;
 
 @end
--- a/CriteoPublisherSdk/Tests/Utility/Checkers/CR_CreativeViewChecker.m
+++ b/CriteoPublisherSdk/Tests/Utility/Checkers/CR_CreativeViewChecker.m
@@ -41,6 +41,19 @@ static NSString *appStoreDisplay = @"https://localhost:9099/display/app-store";
 
 @implementation CR_CreativeViewChecker
 
+- (instancetype)initWithCriteo:(Criteo *)criteo {
+  if (self = [super init]) {
+    [self resetExpectations];
+    _uiWindow = [self createUIWindow];
+    _criteo = criteo;
+    [self resetBannerView];
+    _interstitial = [[CRInterstitial alloc] init];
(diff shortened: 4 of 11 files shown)


CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
```diff
@@ -165,6 +174,11 @@ - (void)loadAdWithBid:(CRBid *)bid {
     return;
   }
 
+  if (self.adUnit.adUnitId == nil && [bid.adUnit isKindOfClass:[CRBannerAdUnit class]]) {
```
TurpIF, line 177: On `self.adUnit.adUnitId == nil` and ``self.adUnit = bid.adUnit`:

Why do we need to register the ad unit of the bid ? I don't see any reason why we should keep track of it. Moreover, this means that you introduce a temporal dependency in `loadAdWithBid` calls.

Given this scenario:
- load successfully bid1 from adUnit1
- display bid1 -> good
- load successfully bid2 from adUnit2
- display bid2 -> not good as the frame same might not match the new ad

Hence I'm not sure that we need to store the adUnit. On the contrary, I think that this make the flow more complicated.

TurpIF, line 177: On `bid.adUnit isKindOf Banner`:

Ok to have such condition because you need the size info. But I would ask: what should we do if it is not the expected kind ? In your dev, in such unexpected case, we just continue displaying and notifying the publisher that everything was good.

From such condition, I expect a `else` branch with a `self.safelyNotifyAdLoadFail:CRErrorCodeNoFill`. And maybe even a log ?

Or you can reuse the `CR_CdbBid` above which contains the size. And to check if this is the expected kind of bid, maybe could you add a method `consumeFor:AdUnitType.Banner` like how it is done [here](https://codesearch.criteois.com/github.com/criteo/android-publisher-sdk/-/blob/publisher-sdk/src/main/java/com/criteo/publisher/Bid.java?L71)

Anyway, as you want to had such kind of logic, could you please add a test case with such scenario:
- Given a successful interstitial bid
- When displaying with a banner view
- Then nothing get displayed
- And listener get notified for failure

CriteoPublisherSdk/Sources/Standalone/CRInterstitial.m
```diff
@@ -184,6 +188,10 @@ - (void)loadAdWithBid:(CRBid *)bid {
     return;
   }
 
+  if (self.adUnit.adUnitId == nil && [bid.adUnit isKindOfClass:[CRInterstitialAdUnit class]]) {
```
TurpIF, line 191: exactly same comment than for banner view. And as you're using such logic twice, I strongly suggest to add an internal consumeFor method in CRBid

CriteoPublisherSdk/Tests/Utility/Checkers/CR_CreativeViewChecker.h
```diff
@@ -36,15 +38,15 @@
 @property(strong, nonatomic, readonly) XCTestExpectation *adCreativeRenderedExpectation;
 @property(strong, nonatomic, readonly) CRBannerView *bannerView;
 @property(strong, nonatomic, readonly) CRInterstitial *interstitial;
-@property(strong, nonatomic, readonly) CRAdUnit *adUnit;
+@property(strong, nonatomic) CRAdUnit *adUnit;
```
TurpIF, line 41: Why can't this stay RO ? Isn't it working by putting an explicit _adUnit = nil in the ctor ?

CriteoPublisherSdk/Tests/Utility/Checkers/CR_CreativeViewChecker.m
```diff
@@ -97,12 +111,15 @@ - (void)injectBidWithExpectedCreativeUrl:(NSString *)creativeUrl
...
 
-- (void)injectBidWithExpectedCreativeUrl:(NSString *)creativeUrl {
-  [self injectBidWithExpectedCreativeUrl:creativeUrl withSkAdNetworkParameters:NO];
+- (void)injectBidWithExpectedCreativeUrl:(NSString *)creativeUrl forAdUnit:(CRAdUnit *)adUnit {
+  [self injectBidWithExpectedCreativeUrl:creativeUrl withSkAdNetworkParameters:NO forAdUnit:adUnit];
 }
 
 - (void)injectBidWithAppStoreClickUrl {
-  [self injectBidWithExpectedCreativeUrl:appStoreDisplay withSkAdNetworkParameters:YES];
+  // FIXME CG
```
TurpIF, line 119: Why this one is a FIXME ?

TurpIF (changes requested):

CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
```diff
@@ -30,14 +30,23 @@ @interface CRBannerView () <WKNavigationDelegate, WKUIDelegate>
...
 @property(nonatomic, readonly) id<CR_URLOpening> urlOpener;
 @property(nonatomic, strong) CR_SKAdNetworkParameters *skAdNetworkParameters;
 
 @end
 
 @implementation CRBannerView
 
+- (instancetype)init {
+  return [self initWithAdUnit:[CRBannerAdUnit alloc]];
+  /*    return [self initWithFrame:CGRectZero
```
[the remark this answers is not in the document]
CedricGatay: I see dead code, removing.

CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
```diff
@@ -30,14 +30,23 @@ @interface CRBannerView () <WKNavigationDelegate, WKUIDelegate>
...
+@property(nonatomic, strong) CRBannerAdUnit *adUnit;
 @property(nonatomic, readonly) id<CR_URLOpening> urlOpener;
 @property(nonatomic, strong) CR_SKAdNetworkParameters *skAdNetworkParameters;
 
 @end
 
 @implementation CRBannerView
 
+- (instancetype)init {
+  return [self initWithAdUnit:[CRBannerAdUnit alloc]];
```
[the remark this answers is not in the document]
CedricGatay: `initWithAdUnit` expects a non nil object, thus allocating an empty one seems clever to me, instead of allowing nil adUnit.
What do you think ?

CriteoPublisherSdk/Tests/Utility/Checkers/CR_CreativeViewChecker.h
CedricGatay: Updated to keep the property RO

CriteoPublisherSdk/Sources/Standalone/CRInterstitial.m
CedricGatay: Added the `consumeFor` method.

CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
CedricGatay: Added the `consumeFor` method and unit tests.

CedricGatay: Remove setting the `adUnit` locally as it is useless.

CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
```diff
@@ -30,14 +30,23 @@ @interface CRBannerView () <WKNavigationDelegate, WKUIDelegate>
 @property(nonatomic) BOOL isResponseValid;
 @property(nonatomic, strong) Criteo *criteo;
 @property(nonatomic, strong) WKWebView *webView;
-@property(nonatomic, readonly) CRBannerAdUnit *adUnit;
+@property(nonatomic, strong) CRBannerAdUnit *adUnit;
```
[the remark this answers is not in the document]
CedricGatay: Made it RO again.

CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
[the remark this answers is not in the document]
TurpIF: Isn't it possible to have the header file declaring a ctor rejecting nil, but the implementation can accept nil ?
It is just that conceptually, nil can be handy to represent the absence of value, but here we have a value but that is empty 😕 
But if it's too boring to get a workaround, ok to leave this.

TurpIF (approved):

CriteoPublisherSdk/Sources/Standalone/CRBannerView.m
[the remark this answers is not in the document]
CedricGatay: We would be declaring conflicting constraints, I'd rather keep the empty thing

PR #243, opened 2022-05-16
fix(DPP-3954): remove MoPub from SDK

CedricGatay:
Remove Mopub from SDK

See [DPP-3954]

*DoD*

* SDK does not use MoPub SDK anymore (no more dependency, no more header bidding)

* SDK does not transmit MoPub consent to CDB

* SDK keep public API for MoPub, they are deprecated and declared to be removed:

** `Criteo.setMoPubConsent`

** `Criteo#Builder.mopubConsent`

* Calling those API should do nothing but log a warning about using a deprecated API

* Changelog is updated to indicate that MoPub is no longer supported


CHANGELOG.md
```diff
@@ -124,7 +124,12 @@ pod clients, you can:
 [gh_releases]: https://github.com/criteo/ios-publisher-sdk/releases
 [use_frameworks]: https://guides.cocoapods.org/syntax/podfile.html#tab_use_frameworks_bang
 
+## Version [NEXT]
+
+### MoPub Adapter
```
TurpIF, line 129: MoPub Header bidding is also removed

CriteoPublisherSdk/Sources/CR_BidManagerHelper.h
```diff
@@ -23,7 +23,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface CR_BidManagerHelper : NSObject
 
-+ (void)removeCriteoBidsFromMoPubRequest:(id)adRequest;
+/**
+ @deprecated Mopub SDK is not supported anymore
+ */
++ (void)removeCriteoBidsFromMoPubRequest:(id)adRequest __deprecated;
```
TurpIF, line 29: This is a private method related to MoPub. Hence you can safely remove it.

CriteoPublisherSdk/Sources/CR_BidManagerHelper.m
```diff
@@ -18,28 +18,12 @@
...
-        if (![pair hasPrefix:@"crt_"]) {
-          [nonCriteoKeywords addObject:pair];
-        }
-      }
-      keywords = [nonCriteoKeywords componentsJoinedByString:@","];
-      [adRequest setValue:keywords forKey:@"keywords"];
-    }
-#pragma clang diagnostic pop
-  }
++ (void)removeCriteoBidsFromMoPubRequest:(id)adRequest __deprecated {
```
TurpIF, line 25: Same than above

CriteoPublisherSdk/Sources/Util/NSObject+Criteo.m
```diff
@@ -54,7 +54,6 @@ + (void)load {
 #pragma ide diagnostic ignored "InfiniteRecursion"
 - (void)cr_loadAd {
```
TurpIF, line 55: This method seems to do nothing now and is part of the hack above that could be removed.

That hack above was swizzling the MPInterstitialAdController#loadAd method to automatically remove the Criteo bids keywords before.
This is part of MoPub support which can be removed now.

I guess the whole `load` override could simply be removed too then.

CriteoPublisherSdk/Tests/IntegrationTests/CRInterstitialDelegateTests.m
```diff
@@ -390,7 +390,7 @@ - (void)testDidFailToReceiveAdContentWithErrorWhenWebViewFailsToNavigate {
   [self cr_waitForExpectations:@[ interstitialWebViewNavigationFailExpectation ]];
 }
 
-- (void)testDidFailToReceiveAdContentWithErrorWebViewFailsToLoad {
+- (void)x_testDidFailToReceiveAdContentWithErrorWebViewFailsToLoad {
```
TurpIF, line 393: why is this test deactivated ?

CriteoPublisherSdk/Tests/UnitTests/CR_CacheManagerTests.m
```diff
@@ -123,6 +124,7 @@ - (void)testSetBidForInterstitial {
   OCMStub(deviceInfo.screenSize).andReturn(CGSizeMake(50, 320));
   [self.cacheManager setBid:self.testBid];
   XCTAssertTrue([[self.cacheManager getBidForAdUnit:self.adUnit] isEqual:self.testBid]);
+  Criteo.sharedCriteo.dependencyProvider.deviceInfo = _originDeviceInfo;
```
TurpIF, line 127: Each method that are calling `mockDeviceInfo` should then restore the `dependencyProvider.deviceInfo`. What about putting this restoration in a `tearDown` method so that tests don't need to clean up internal states ?

Also, as the mock is a partial mock, it is fine to always have it. So why not setting the partial mock within the `setUp` method directly ? So all new tests could easily stub/verify the behavior:

```
@property originDeviceInfo;
@property spyDeviceInfo;

- setUp {
 _originDeviceInfo = dependencyProvider.deviceInfo;
 _spyDeviceInfo = OCMPartialMock(_originDeviceInfo);
 dependencyProvider.deviceInfo = _spyDeviceInfo;
}

- tearDown {
  dependencyProvider.deviceInfo = _originDeviceInfo;
}

- testFoo {
  OCMStub(self.spyDeviceInfo.screenSize).andReturn(...);
  ...
}
```

TurpIF (changes requested):

PR #264, opened 2023-01-20
DPP-4285: crash fix

vapopa:
handle exception on body serialization for POST request

### DIFF
--- a/CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m
+++ b/CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m
@@ -173,14 +173,14 @@
 }
 
 - (void)testNetworkManagerFailedSerialization {
-    NSURL *url;
-    NSData *body;
-    [self.networkManager postToUrl:url
-                              body:body
-                   responseHandler:^(NSData * _Nullable data, NSError * _Nullable error) {
-        XCTAssertNotNil(error);
-        XCTAssertNil(data);
-    }];
+  NSURL *url;
+  NSData *body;
+  [self.networkManager postToUrl:url
+                            body:body
+                 responseHandler:^(NSData *_Nullable data, NSError *_Nullable error) {
+                   XCTAssertNotNil(error);
+                   XCTAssertNil(data);
+                 }];
 }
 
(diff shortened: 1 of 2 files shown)


TurpIF (commented):
What about adding a test on this change ?

CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m
```diff
@@ -172,6 +172,17 @@ - (void)testNetworkManagerPostCall {
                               }]]);
 }
 
+- (void)testNetworkManagerFailedSerialization {
+    NSURL *url;
+    NSData *body;
+    [self.networkManager postToUrl:url
+                              body:body
+                   responseHandler:^(NSData * _Nullable data, NSError * _Nullable error) {
+        XCTAssertNotNil(error);
```
TurpIF: If the networkManager is not calling the response handler, then this test could easily pass. Maybe could you add a synchronization point to ensure that the response handler gets invoked ?

See this other test for instance: https://github.com/criteo/ios-publisher-sdk/pull/264/commits/e7a28c15131fa08f885eebd2f776c660412046c8#diff-60683f30ec672a62f5f5605e46eeb74824380a69dc3e3accea852459d77ce1aeR60

CriteoPublisherSdk/Tests/UnitTests/Network/CR_NetworkManagerTests.m
```diff
@@ -173,14 +173,14 @@ - (void)testNetworkManagerPostCall {
...
-                   responseHandler:^(NSData * _Nullable data, NSError * _Nullable error) {
-        XCTAssertNotNil(error);
-        XCTAssertNil(data);
-    }];
+  NSURL *url;
+  NSData *body;
+  [self.networkManager postToUrl:url
+                            body:body
+                 responseHandler:^(NSData *_Nullable data, NSError *_Nullable error) {
+                   XCTAssertNotNil(error);
```
TurpIF, line 181: Why not fixing the initial commit by amending it rather than adding many fixup commits which could add noise when digging in the repository ? Also it make things harder to be reverted as you'll need to revert many commit at once.

PR #345, opened 2025-01-07
[DRAFT] [ADD] Swift Package Manager Support

### LINKED ISSUE #94: Swift Package Manager Support
Hello 👋 do you plan to support the [Swift Package Manager](https://swift.org/package-manager/)?

mflknr:
Hello,

this PR tries to add SPM support in addition to CocoaPods since these are in maintenance mode. The source code compiles and works (as far as i can tell), but I hit a few roadblocks where I figured to ask for some guidance.

1. The public headers: The podspecs defines the "Public" folder as the path where the public headers are located. Yet, in libraries were Criteo is used with pods, i observed that ALL headers are available and therefor public. The package file seems to be more strict and only makes the header in the "Public" folder public. Thats why I initially said in #94 that I cannot use the xcframework binary because the headers are not complete. They are complete, but just differently available over pods. It is working now, but I need feedback if everything is semantically correct.
2. Tests: I had to move some files and therefor targets back and forth because SPM does not support swift/objc mixed targets (which was really bad tbh). The original testplan are no longer working and I cant figure out what was tested there. So someone has recreate the testplans from scratch. We also run into the same problem with the public headers here ...
3. Automation: Without working testplans, I cant even try to setup fastlane again. So ... there is that.

Anyway, I really need feedback here to continue on this. Thanks.

Closes #94


mflknr:
Just to get any life signal:
@leonardlabat 
@vguerci 
@afewcc 
@dzhang-criteo 

Anyone? Is this even still maintained?

acomets:
Hi @mflknr,

Sorry for the long delay but stopped maintaining the Criteo SDK and we will be sunsetting it by end of September so there is no point adding SPM support.

Thanks for your contribution!

Antoine

mflknr:
Hello @acomets and thank you for the reply. What are the alternatives of using this SDK?

